Issue #6, opened 2025-06-05
Deploy to Maven Central.


Anmol202005:
Deploy the project to Maven Central Repository to enable easy integration with tools like Checkstyle and other Java projects.

Once deployed other projects can integrate Autofix using plugin.

rdiachenko:
Please check commits of how it was done in another project - https://github.com/checkstyle/patch-filters/issues/306, we need to do the same

kangcaleb:
Hi! Do yall know when this is going to be deployed to Maven Central? The maven instructions here are not working for me :(

https://docs.openrewrite.org/running-recipes/popular-recipe-guides/automatically-fix-checkstyle-violations

romani:
@rdiachenko , @sabaka , @Anmol202005 , @MonuChaudhary14 , I think I need to help here to scaffold configs for deploy to let me place variables in secrets and hope it will work.

romani:
done by @stoyanK7 at https://github.com/checkstyle/checkstyle-openrewrite-recipes/commit/922426c99517b6119d32a24ddb15d8df0595e1f3

org secrets are setup https://github.com/organizations/checkstyle/settings/secrets/actions

released by 
https://github.com/checkstyle/checkstyle-openrewrite-recipes/actions/workflows/release.yml

first release 1.0.0
https://github.com/checkstyle/checkstyle-openrewrite-recipes/actions/runs/32309507971/job/96253253705

auto publishing, but can be watched at: https://central.sonatype.com/publishing/deployments

done:
https://repo1.maven.org/maven2/com/puppycrawl/tools/checkstyle-openrewrite-recipes/1.0.0/

PR #9, opened 2025-06-08
Deployed to Maven Central

Anmol202005:
Fixes: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/6
Configured pom.xml for maven central deployment.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -5,7 +5,13 @@
                              http://maven.apache.org/xsd/maven-4.0.0.xsd">
     <modelVersion>4.0.0</modelVersion>
 
-    <groupId>org.checkstyle.autofix</groupId>
+    <parent>
+        <groupId>org.sonatype.oss</groupId>
+        <artifactId>oss-parent</artifactId>
+        <version>9</version>
+    </parent>
+
+    <groupId>com.puppycrawl.tools</groupId>
     <artifactId>checkstyle-openrewrite-recipes</artifactId>
     <version>1.0.0-SNAPSHOT</version>
     <packaging>jar</packaging>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -5,6 +5,13 @@
...
     <modelVersion>4.0.0</modelVersion>
 
+    <!-- This parent is requirements for deploy of artifacts to maven central -->
+    <parent>
+        <groupId>org.sonatype.oss</groupId>
+        <artifactId>oss-parent</artifactId>
+        <version>9</version>
+    </parent>
+
     <groupId>org.checkstyle.autofix</groupId>
```
romani, line 15: I think we can `org.checkstyle` only for group.
it is ok to have plugins/tools to be just under "checkstyle"

rdiachenko: In the patch-filter project I see latest master has `<groupId>com.puppycrawl.tools</groupId>`, shall we go with the same groupId? Better to agree on this now while we are still at the start of the project

romani: https://repo1.maven.org/maven2/org/checkstyle

Is not existing, so it means it will be a special process to claim ownership of such domain.

Lets use group common for all , as Ruslan suggested 
https://repo1.maven.org/maven2/com/puppycrawl/tools/

rdiachenko:
@Anmol202005 please resolve conflicts

rdiachenko (approved):
The change looks good to me. @romani please take a look when you have time as you have a better context regarding deployment to maven central

romani (approved):
lets go.

I will do release only after we make sure it works on SNAPSHOT version

PR #19, opened 2025-06-20
added license header

Anmol202005:
Fixes: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/17
Added License Header to all the source and test files.

### DIFF
--- a/config/checkstyle.properties
+++ b/config/checkstyle.properties
@@ -1,8 +1,7 @@
-
 checkstyle.suppressions.file=config/suppressions.xml
 checkstyle.suppressions-xpath.file=config/suppressions.xml
-checkstyle.header.file=https://raw.githubusercontent.com/checkstyle/checkstyle/master/config/java.header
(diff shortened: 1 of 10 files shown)


Anmol202005:
@rdiachenko removed package-info suppression too.

romani:
consider to think on auto fix for this Check.

config/regexp-header.txt  is regexp, not acualtext block to insert.
Auto fix (Recipie) for such Checks should have some config to profile path to file that will have ready to use block of text to insert into file with violation.

rdiachenko:
> update is good.
> 
> consider to think on auto fix for this Check.
> 
> config/regexp-header.txt is regexp, not acualtext block to insert. Auto fix (Recipie) for such Checks should have some config to profile path to file that will have ready to use block of text to insert into file with violation.

And this is a great next cadnidate for a recipe that is based on parameterized check - https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/25

Issue #14, opened 2025-06-11
Add Meta-Data for recipies.


Anmol202005:
Follow up of: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/1

The src/main/resources/META-INF/rewrite/checkstyle.yml file is used for OpenRewrite recipe discovery in the project.

Add meta-data for the existing recipes in `recipes.yml`.
This file contains declarations of OpenRewrite recipes that can automatically refactor and modernize your codebase.

rdiachenko:
We need to find a way to have name and description in a single place, some context is here: https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/20#discussion_r2159624816

Anmol202005:
>It will be hard to keep consistency if we have multiple places for same values.

@rdiachenko  since `.yml` file is used to configure rewrite. can we close this !!

PR #20, opened 2025-06-20
added META data

Anmol202005:
Fixes: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/14
Added meta-data for the existing recipes in recipes.yml.

### DIFF
--- a/src/main/resources/META-INF/rewrite/recipes.yml
+++ b/src/main/resources/META-INF/rewrite/recipes.yml
@@ -1,0 +1,10 @@
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.checkstyle.autofix.CheckstyleAutoFix
+displayName: Checkstyle autoFix
+description: Automatically fixes Checkstyle violations.
+tags:
+  - checkstyle
+  - static-analysis
+recipeList:
+  - org.checkstyle.autofix.recipe.UpperEllRecipe


src/main/resources/META-INF/rewrite/recipes.yml
```diff
@@ -0,0 +1,7 @@
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.checkstyle.autofix.CheckstyleAutoFix
+displayName: Checkstyle autoFix
+description: Automatically fixes Checkstyle violations.
```
rdiachenko, line 5: Should we add tags for better discoverability just for the top most recipe?
```
tags:
    - checkstyle
    - static-analysis 
```

src/main/resources/META-INF/rewrite/recipes.yml
```diff
@@ -0,0 +1,7 @@
+---
+type: specs.openrewrite.org/v1beta/recipe
+name: org.checkstyle.autofix.CheckstyleAutoFix
+displayName: Checkstyle autoFix
```
rdiachenko, line 4: There are `getDisplayName()` and `getDescription()` overriden in code with same values. It will be hard to keep consistency if we have multiple places for same values. Could you check what's OpenRewrite fallback behaviour? If we describe metadata in `recipes.yml` only, can we remove corresponding overrides from code and use `recipes.yml` as the source of truth?

Anmol202005: We cannot remove `getDisplayName()` and `getDescription()` - these are abstract methods in the Recipe class that must be implemented.

However,  we can add tags directly to `recipes.yml` file.
OpenRewrite will automatically use those tags when the recipe is loaded via YAML configuration.

src/main/resources/META-INF/rewrite/recipes.yml
Anmol202005: done.

rdiachenko (approved):
lgtm

Issue #23, opened 2025-06-22
Update UpperEll to use Checkstyle violations as a context


rdiachenko:
Currently, `UpperEll` recipe fixes all discovered cases. But we want it to fix only those cases where Checkstyle produces violaitons for.

Update `UpperEll` recipe to match source file and line number from the violation and fix it only when there is a match.

As OpenRewrite doesn't give position for the node in the tree, we need to solve this outselves. Something like this should cover most of the cases:
```java
int literalIndex = fullSource.indexOf(literalText);
            
if (literalIndex >= 0) {
  textBeforeLiteral = fullSource.substring(0, literalIndex);
}
// count new lines
textBeforeLiteral.chars().filter(ch -> ch == '\n').count();
```

rdiachenko:
Resolved by https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/27

PR #27, opened 2025-06-23
Updated UpperEll to use Checkstyle violations

Anmol202005:
Fixes: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/23
Updated UpperEll recipe to match source file and line number from the violation and fix it only when there is a match.

### DIFF
--- a/src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
+++ b/src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
@@ -17,9 +17,19 @@
 
 package org.checkstyle.autofix.recipe;
 
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CancellationException;
+import java.util.function.Function;
+
+import org.checkstyle.autofix.parser.CheckstyleViolation;
+import org.openrewrite.Cursor;
 import org.openrewrite.ExecutionContext;
+import org.openrewrite.PrintOutputCapture;
 import org.openrewrite.Recipe;
 import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.RecipeRunException;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaType;
@@ -29,6 +39,16 @@ import org.openrewrite.java.tree.JavaType;
  * in long literals with uppercase 'L'.
  */
 public class UpperEll extends Recipe {
+
+    private final List<CheckstyleViolation> violations;
+
+    public UpperEll() {
+        this(new ArrayList<>());
+    }
+
+    public UpperEll(List<CheckstyleViolation> violations) {
+        this.violations = violations;
+    }
 
     @Override
     public String getDisplayName() {
@@ -46,20 +66,121 @@ public class UpperEll extends Recipe {
         return new UpperEllVisitor();
     }
 
-    /**
-     * Visitor that replaces lowercase 'l' suffixes in long literals with uppercase 'L'.
-     */
-    private static final class UpperEllVisitor extends JavaIsoVisitor<ExecutionContext> {
+    private final class UpperEllVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        private static final String LOWERCASE_L = "l";
+        private static final String UPPERCASE_L = "L";
+
+        private Path sourcePath;
+
+        @Override
+        public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
+            this.sourcePath = cu.getSourcePath();
+            return super.visitCompilationUnit(cu, ctx);
+        }
+
         @Override
         public J.Literal visitLiteral(J.Literal literal, ExecutionContext ctx) {
             J.Literal result = super.visitLiteral(literal, ctx);
             final String valueSource = result.getValueSource();
 
-            if (valueSource != null && valueSource.endsWith("l")
-                    && result.getType() == JavaType.Primitive.Long) {
+            if (valueSource != null && valueSource.endsWith(LOWERCASE_L)
+                    && result.getType() == JavaType.Primitive.Long
+                    && isAtViolationLocation(result)) {
+
                 final String numericPart = valueSource.substring(0, valueSource.length() - 1);
-                final String newValueSource = numericPart + "L";
-                result = result.withValueSource(newValueSource);
+                result = result.withValueSource(numericPart + UPPERCASE_L);
+            }
+
+            return result;
+        }
+
+        private boolean isAtViolationLocation(J.Literal literal) {
+            final J.CompilationUnit cursor = getCursor().firstEnclosing(J.CompilationUnit.class);
+
+            final int line = computeLinePosition(cursor, literal, getCursor());
+            final int column = computeColumnPosition(cursor, literal, getCursor());
+
+            return violations.stream().anyMatch(violation -> {
+                return violation.getLine() == line
+                        && violation.getColumn() == column
+                        && Path.of(violation.getFileName()).equals(sourcePath);
+            });
+        }
+
+        /**
+         * Computes the position of a target element within a syntax tree using position calculator.
+         * This method traverses the given syntax tree and captures the printed output until the
+         * target element is encountered. When the target is found, a CancellationException
+         * is thrown to interrupt traversal, and the captured output is passed to the provided
+         * positionCalculator to compute the position.
+         *
+         * @param tree the root of the syntax tree to traverse
+         * @param targetElement the element whose position is to be computed
+         * @param cursor the current cursor in the tree traversal
+         * @param positionCalculator a function to compute the position from the printed output
+         * @return the computed position of the target element
+         * @throws IllegalStateException if the target element is not found in the tree
+         * @throws RecipeRunException if an error occurs during traversal
+         */
+        private int computePosition(
+                J tree,
+                J targetElement,
+                Cursor cursor,
+                Function<String, Integer> positionCalculator
+        ) {
+            final TreeVisitor<?, PrintOutputCapture<TreeVisitor<?, ?>>> printer =
+                    tree.printer(cursor);
+
+            final PrintOutputCapture<TreeVisitor<?, ?>> capture =
+                    new PrintOutputCapture<>(printer) {
+                        @Override
+                        public PrintOutputCapture<TreeVisitor<?, ?>> append(String text) {
+                            if (targetElement.isScope(getContext().getCursor().getValue())) {
+                                super.append(targetElement.getPrefix().getWhitespace());
+                                throw new CancellationException();
+                            }
+                            return super.append(text);
+                        }
+                    };
+
+            final int result;
+            try {
+                printer.visit(tree, capture, cursor.getParentOrThrow());
+                throw new IllegalStateException("Target element: " + targetElement
+                        + ", not found in the syntax tree.");
+            }
+            catch (CancellationException exception) {
+                result = positionCalculator.apply(capture.getOut());
+            }
+            catch (RecipeRunException exception) {
+                if (exception.getCause() instanceof CancellationException) {
+                    result = positionCalculator.apply(capture.getOut());
+                }
+                else {
+                    throw exception;
+                }
+            }
+            return result;
+        }
+
+        private int computeLinePosition(J tree, J targetElement, Cursor cursor) {
+            return computePosition(tree, targetElement, cursor,
+                    out -> 1 + Math.toIntExact(out.chars().filter(chr -> chr == '\n').count()));
+        }
+
+        private int computeColumnPosition(J tree, J targetElement, Cursor cursor) {
+            return computePosition(tree, targetElement, cursor, this::calculateColumnOffset);
+        }
+
+        private int calculateColumnOffset(String out) {
+            final int lineBreakIndex = out.lastIndexOf('\n');
+            final int result;
+            if (lineBreakIndex == -1) {
+                result = out.length();
+            }
+            else {
+                result = out.length() - lineBreakIndex - 1;
             }
             return result;
         }
--- a/src/test/java/org/checkstyle/autofix/recipe/UpperEllTest.java
+++ b/src/test/java/org/checkstyle/autofix/recipe/UpperEllTest.java
@@ -18,7 +18,11 @@
 package org.checkstyle.autofix.recipe;
 
 import java.io.IOException;
+import java.nio.file.Path;
+import java.util.List;
 
+import org.checkstyle.autofix.parser.CheckstyleReportsParser;
+import org.checkstyle.autofix.parser.CheckstyleViolation;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.Recipe;
 
@@ -26,7 +30,12 @@ public class UpperEllTest extends AbstractRecipeTest {
 
     @Override
     protected Recipe getRecipe() {
-        return new UpperEll();
+        final String reportPath = "src/test/resources/org/checkstyle/autofix/recipe/upperell"
+                + "/report.xml";
+
+        final List<CheckstyleViolation> violations =
+                CheckstyleReportsParser.parse(Path.of(reportPath));
+        return new UpperEll(violations);
     }
 
     @Test
--- a/src/test/resources/org/checkstyle/autofix/recipe/upperell/hexoctalliteral/InputHexOctalLiteral.java
+++ b/src/test/resources/org/checkstyle/autofix/recipe/upperell/hexoctalliteral/InputHexOctalLiteral.java
@@ -2,14 +2,14 @@ package org.checkstyle.autofix.recipe.upperell.hexoctalliteral;
 
 public class InputHexOctalLiteral {
     private long hexLower = 0x1ABCl;
-        private long hexUpper = 0X2DEFl;
-        private long octal = 0777l;
-        private long binary = 0b1010l;
-        private long decimal = 12345l;
+    private long hexUpper = 0X2DEFl;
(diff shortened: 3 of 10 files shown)


Anmol202005:
@rdiachenko 
please review.

rdiachenko (commented):
@Anmol202005 could you add test cases for when some lines are suppressed by checkstyle (so they are not in the report) we don't change those lines? Same for column case, having 2 violations on the same line but one violation is suppressed, so we need to fix only one.

src/test/resources/org/checkstyle/autofix/recipe/upperell/hexoctalliteral/InputHexOctalLiteral.java
```diff
@@ -1,15 +1,15 @@
...
-    private long hexLower = 0x1ABCl;
-        private long hexUpper = 0X2DEFl;
-        private long octal = 0777l;
-        private long binary = 0b1010l;
-        private long decimal = 12345l;
+    private long hexLower = 0x1ABCL;
+    private long hexUpper = 0X2DEFL;
+    private long octal = 0777L;
+    private long binary = 0b1010L;
+    private long decimal = 12345L;
```
rdiachenko, line 8: Why did you change all to upper 'l'?

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
```diff
@@ -43,23 +64,112 @@ public String getDescription() {
...
+        private final List<CheckstyleViolation> violations;
+        private String currentFileName;
+
+        UpperEllVisitor(List<CheckstyleViolation> violations) {
+            this.violations = violations;
+        }
+
+        @Override
+        public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
+            this.currentFileName = cu.getSourcePath().toString();
```
rdiachenko, line 81: `toString()` may not be reliable, please check other options

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
```diff
@@ -43,23 +64,112 @@ public String getDescription() {
...
+                    tree.printer(cursor);
+
+            final PrintOutputCapture<TreeVisitor<?, ?>> capture =
+                    new PrintOutputCapture<>(printer) {
+                        @Override
+                        public PrintOutputCapture<TreeVisitor<?, ?>> append(String text) {
+                            if (targetElement.isScope(getContext().getCursor().getValue())) {
+                                targetElement.getPrefix();
+                                super.append(targetElement.getPrefix().getWhitespace());
+                                throw new CancellationException();
```
rdiachenko, line 131: why do we need this exception? it is from concurrent package, which is not related to the current logic

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
```diff
@@ -43,23 +64,112 @@ public String getDescription() {
...
+        ) {
+            final TreeVisitor<?, PrintOutputCapture<TreeVisitor<?, ?>>> printer =
+                    tree.printer(cursor);
+
+            final PrintOutputCapture<TreeVisitor<?, ?>> capture =
+                    new PrintOutputCapture<>(printer) {
+                        @Override
+                        public PrintOutputCapture<TreeVisitor<?, ?>> append(String text) {
+                            if (targetElement.isScope(getContext().getCursor().getValue())) {
+                                targetElement.getPrefix();
```
rdiachenko, line 129: what this line is about? you don't even use its result

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
```diff
@@ -43,23 +64,112 @@ public String getDescription() {
...
+            }
+            catch (RecipeRunException exception) {
+                if (exception.getCause() instanceof CancellationException) {
+                    result = positionCalculator.apply(capture.getOut());
+                }
+                else {
+                    throw exception;
+                }
+            }
+            return result;
```
rdiachenko, line 153: I don't understand this logic, why do we need this? please try to simplify

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
```diff
@@ -43,23 +64,112 @@ public String getDescription() {
...
-                final String newValueSource = numericPart + "L";
-                result = result.withValueSource(newValueSource);
+                result = result.withValueSource(numericPart + "L");
+            }
+
+            return result;
+        }
+
+        private boolean isAtViolationLocation(J.Literal literal) {
+            final J.CompilationUnit cursor = Objects.requireNonNull(getCursor()
```
rdiachenko, line 102: why do you need `Objects.requireNonNull` here?

rdiachenko (changes requested):

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
Anmol202005: The one line fix i mentioned on discord.

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
Anmol202005: :)

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
Anmol202005: Tried other options none seems reliable.
IMO `toString()` is reliable.

`cu.getSourcePath()` returns a `Path`, so this should be good.

if i am missing something please suggest other option.

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
Anmol202005: added while debugging will remove.

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
Anmol202005: This handles two ways the `CancellationException` can come back:

Direct catch - when their `append()` method throws `CancellationException` directly
Wrapped catch - when the visitor framework wraps their exception inside a `RecipeRunException`

Some visitor implementations catch exceptions and re-wrap them, so they need to handle both the original exception and the wrapped version. If no exception occurs at all, it means the target wasn't found.

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
Anmol202005: done. removed.

src/test/resources/org/checkstyle/autofix/recipe/upperell/hexoctalliteral/InputHexOctalLiteral.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
```diff
@@ -43,23 +63,108 @@ public String getDescription() {
...
+        }
+
+        private boolean isAtViolationLocation(J.Literal literal) {
+            final J.CompilationUnit cursor = getCursor().firstEnclosing(J.CompilationUnit.class);
+
+            final int line = computeLinePosition(cursor, literal, getCursor());
+            final int column = computeColumnPosition(cursor, literal, getCursor());
+
+            return violations.stream().anyMatch(violation -> {
+                return violation.getLine() == line
```
timurt, line 107: `return` control statement is redundant here

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
```diff
@@ -43,23 +63,108 @@ public String getDescription() {
...
+                    out -> 1 + (int) out.chars().filter(chr -> chr == '\n').count());
+        }
+
+        private int computeColumnPosition(J tree, J targetElement, Cursor cursor) {
+            return computePosition(tree, targetElement, cursor, this::calculateColumnOffset);
+        }
+
+        private int calculateColumnOffset(String out) {
+            final int lineBreakIndex = out.lastIndexOf('\n');
+            final int result;
```
timurt, line 162: can it be simplified to 

```
return lineBreakIndex == -1 ? out.length() : out.length() - lineBreakIndex - 1;
```

?

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
```diff
@@ -43,23 +63,108 @@ public String getDescription() {
...
+            final int column = computeColumnPosition(cursor, literal, getCursor());
+
+            return violations.stream().anyMatch(violation -> {
+                return violation.getLine() == line
+                        && violation.getColumn() == column
+                        && Path.of(violation.getFileName()).equals(Path.of(currentFileName));
+            });
+        }
+
+        private int computePosition(
```
timurt, line 113: can you add javadoc, the method looks complex

smth like, feel free to modify

```
         * Computes the position of a target element within a tree by traversing
         * the tree and capturing output until the target is found.
         * 
         * <p>This method uses a custom PrintOutputCapture that throws a
         * CancellationException when the target element is encountered, allowing
         * us to determine its position in the output.</p>
```

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
```diff
@@ -43,23 +63,108 @@ public String getDescription() {
...
+            }
+            catch (CancellationException | RecipeRunException exception) {
+                if (exception instanceof CancellationException
+                        || exception.getCause() instanceof CancellationException) {
+                    result = positionCalculator.apply(capture.getOut());
+                }
+                else {
+                    throw exception;
+                }
+            }
```
timurt, line 147: nit: it is up to you, but splitting exceptions might bring more readability, feel free to ignore me :) 

```
            ...
            } catch (CancellationException e) {
                return positionCalculator.apply(capture.getOut());
            } catch (RecipeRunException e) {
                if (e.getCause() instanceof CancellationException) {
                    return positionCalculator.apply(capture.getOut());
                }
                throw e;
            }
```

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
Anmol202005: @timurt Checkstyle check blocks this.

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
Anmol202005:
```java
return violations.stream().anyMatch(violation -> violation.getLine() == line
                    && violation.getColumn() == column
                    && Path.of(violation.getFileName()).equals(Path.of(currentFileName)));
        }
```

this causes checkstyle violation:
```bash
[ERROR] src/main/java/org/checkstyle/autofix/recipe/UpperEll.java:[106,59] (blocks) NeedBraces: '->' construct must use '{}'s.
```

timurt: so sad :( 

okay, no worries

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
timurt: -_-

I won't fight checkstyle rules here

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
Anmol202005: done.

timurt (approved):
Well done!

src/test/java/org/checkstyle/autofix/recipe/UpperEllTest.java
```diff
@@ -18,29 +18,40 @@
...
+import org.checkstyle.autofix.parser.CheckstyleViolation;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.Recipe;
 
 public class UpperEllTest extends AbstractRecipeTest {
 
     @Override
-    protected Recipe getRecipe() {
-        return new UpperEll();
+    protected Recipe getRecipe() throws XMLStreamException, IOException {
```
rdiachenko, line 34: Let's catch both `XMLStreamException, IOException` and rethrow `IllegalArgumentException` from within `CheckstyleReportsParser.parse`, this will allow to keep tests cleaner and fail fast when provided report has issues

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
```diff
@@ -30,6 +40,16 @@
  */
 public class UpperEll extends Recipe {
 
+    private List<CheckstyleViolation> violations;
```
rdiachenko, line 43: make it `final`

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
```diff
@@ -30,6 +40,16 @@
...
+
+    public UpperEll() {
+        this.violations = new ArrayList<>();
+    }
```
rdiachenko, line 47: Why do we need this default constructor? If we do need it, let's update it to `this(new ArrayList<>());`

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
```diff
@@ -43,23 +63,128 @@ public String getDescription() {
...
 
-    /**
-     * Visitor that replaces lowercase 'l' suffixes in long literals with uppercase 'L'.
-     */
     private static final class UpperEllVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        private static final String LOWERCASE_L = "l";
+        private static final String UPPERCASE_L = "L";
+
+        private final List<CheckstyleViolation> violations;
```
rdiachenko, line 74: Can we reuse this list from the outer class? This will require making `UpperEllVisitor` non-static, but this is ok to avoid such duplication

src/main/java/org/checkstyle/autofix/recipe/UpperEllRecipe.java
rdiachenko: So, here you get a `Path`, convert it to `String` and during usage:
```
Path.of(violation.getFileName()).equals(Path.of(currentFileName))
```
you convert `String` back to `Path`. So why not to keep it as `Path` initially?

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
```diff
@@ -43,23 +63,128 @@ public String getDescription() {
...
+                else {
+                    throw exception;
+                }
+            }
+            return result;
+        }
+
+        private int computeLinePosition(J tree, J targetElement, Cursor cursor) {
+            return computePosition(tree, targetElement, cursor,
+                    out -> 1 + (int) out.chars().filter(chr -> chr == '\n').count());
```
rdiachenko, line 173: Instead of casting it to `(int)`, let's use `Math.toIntExact`

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
```diff
@@ -43,23 +63,128 @@ public String getDescription() {
...
+                                throw new CancellationException();
+                            }
+                            return super.append(text);
+                        }
+                    };
+
+            final int result;
+            try {
+                printer.visit(tree, capture, cursor.getParentOrThrow());
+                throw new IllegalStateException("Target element not found in tree");
```
rdiachenko, line 155: When you get this message, it is not clear which element caused it. Let's include the target element, so it is easier to debug

Anmol202005: done.

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
Anmol202005: done.

Anmol202005:
@rdiachenko done.

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
```diff
@@ -46,20 +66,121 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
-    /**
-     * Visitor that replaces lowercase 'l' suffixes in long literals with uppercase 'L'.
-     */
-    private static final class UpperEllVisitor extends JavaIsoVisitor<ExecutionContext> {
+    private final class UpperEllVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        private static final String LOWERCASE_L = "l";
+        private static final String UPPERCASE_L = "L";
+
+        private Path currentFileName;
```
rdiachenko, line 74: This is a path, not a file name. Let's rename it to `sourcePath`

src/test/java/org/checkstyle/autofix/recipe/UpperEllTest.java
rdiachenko: We don't need `throws XMLStreamException, IOException`, please remove it from all tests

src/main/java/org/checkstyle/autofix/recipe/UpperEll.java
Anmol202005: done.

rdiachenko (approved):
lgtm

Issue #2, opened 2025-06-03
Analyze all Checkstyle checks for OpenRewrite autofix feasibility.


Anmol202005:
### Goal
Create a comprehensive analysis of all Checkstyle checks to determine which violations can be automatically fixed with OpenRewrite recipes and which require manual intervention.

### Tasks
-  Inventory all Checkstyle checks by category
-  Classify each check as:
     -  Autofixable
     -  Not Autofixable
- Document reasoning for each classification
- Create priority list for recipe development
- Update status as recipes are implemented

rdiachenko:
@Anmol202005 please suggest next check that has some configurable parameters. This will be the next candidate for autofix when we finish with UpperEll end-to-end flow.

romani:
@Anmol202005 , I think you did update for README.md already to do this, 

do we have some leftover ?  please mention it, or lets close issue.

MonuChaudhary14:
@romani 

Can we close this issue?

PR #34, opened 2025-07-02
Updated README.md

Anmol202005:
Part of #2:
Updated README.md added:

- AbbreviationAsWordInName
- AbstractClassName
- UnusedLocalVariable

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,10 +34,11 @@ _No checks analyzed yet_
 
 ### Coding
 
-| Status | Check                                                                                                              | Recipe | Coverage Notes                                                                           |
-|--------|--------------------------------------------------------------------------------------------------------------------|--------|------------------------------------------------------------------------------------------|
-| 🟢     | [`FinalLocalVariable`](https://checkstyle.sourceforge.io/checks/coding/finallocalvariable.html#FinalLocalVariable) | `TBD`  |                                                                                          |
-| 🔴     | [`MagicNumber`](https://checkstyle.sourceforge.io/checks/coding/magicnumber.html#MagicNumber)                      |        | it requires contextual understanding to replace literals with meaningful named constants |
+| Status | Check                                                                                                                  | Recipe | Coverage Notes                                                                           |
+|--------|------------------------------------------------------------------------------------------------------------------------|--------|------------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -19,7 +19,10 @@ This table tracks the auto-fix support status of OpenRewrite recipes for each Ch
 
 ### Annotations
 
-_No checks analyzed yet_
+| Status | Check                                                                                                                                | Recipe | Coverage Notes                          |
+|--------|--------------------------------------------------------------------------------------------------------------------------------------|--------|-----------------------------------------|
+| 🟢     | [`AbbreviationAsWordInName`](https://checkstyle.sourceforge.io/checks/naming/abbreviationaswordinname.html#AbbreviationAsWordInName) | `TBD`  |                                         |
```
romani, line 24: I don't know how it can be auto fix?
We don't know what user meant

README.md
```diff
@@ -19,7 +19,10 @@ This table tracks the auto-fix support status of OpenRewrite recipes for each Ch
...
+| 🟢     | [`AbstractClassName`](https://checkstyle.sourceforge.io/checks/naming/abstractclassname.html#AbstractClassName)                      | `TBD`  |                                         |
```
romani, line 25: I don't know how it can be auto fix?
We don't know what user meant .
It has two options to fix .

romani (commented):
Thoughts

README.md
Anmol202005: in cases like :
`abstract class XXX {}`
we can change the class name to `AbstractXXX`.

but if we change the class name then we also need to update its usage too.

Anmol202005:
@romani updated both to `Won't Be Covered`.

rdiachenko:
@Anmol202005 let's put all such commits under this issue https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/2

Anmol202005:
> @Anmol202005 let's put all such commits under this issue #2

@rdiachenko done.

README.md
rdiachenko: The check uses name pattern to check and other two properties to control when to check. I think we can try to cover it by having a recipe that accepts same pattern and does renaming according to the pattern. It won't be fully covered, but we can try to cover the check partially at least.

rdiachenko: @Anmol202005 let's change it to "Partial Coverag" for now, we can always set "not covered" if nothing we can do

Anmol202005: done.

Anmol202005:
@rdiachenko Done.

Issue #36, opened 2025-07-05
load checkstyle config from config file uri


Anmol202005:
Update `CheckstyleAutoFix.java` and implement loading checkstyle-config.

comment: https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/35#discussion_r2188433518

Create `Configuration` class similar to `com.puppycrawl.tools.checkstyle.api.Configuration`

rdiachenko:
Resolved by https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/39

PR #39, opened 2025-07-07
Issue #36: Updated CheckstyleAutoFix to load checkstyle-config.

Anmol202005:
Updated Updated CheckstyleAutoFix to load checkstyle-config.

Implemented property configuration as applied in checkstyle `Main.java` (https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L370).

Created `Configuration` Class to store checkstyle configuration.


src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
```diff
@@ -49,12 +69,38 @@ public String getViolationReportPath() {
...
+        }
+        else {
+            try (FileInputStream input = new FileInputStream(propFile)) {
+                props.load(input);
+            }
+            catch (FileNotFoundException exception) {
+                throw new IllegalArgumentException("Failed to read: " + propFile, exception);
+            }
+        }
+        return (Configuration) ConfigurationLoader.loadConfiguration(
```
rdiachenko, line 103: you are trying to case internal checkstyle class to your own, I think you will get runtime exception here when you try to run it

src/main/java/org/checkstyle/autofix/parser/Configuration.java
```diff
@@ -0,0 +1,41 @@
...
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package org.checkstyle.autofix.parser;
+
+import java.io.Serializable;
+import java.util.Map;
+
+import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
+
+public interface Configuration extends Serializable {
```
rdiachenko, line 25: we don't need `Serializable`

src/main/java/org/checkstyle/autofix/parser/Configuration.java
```diff
@@ -0,0 +1,41 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package org.checkstyle.autofix.parser;
+
+import java.io.Serializable;
+import java.util.Map;
+
+import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
```
rdiachenko, line 23: please, no checkstyle dependences in our custom model classes

src/main/java/org/checkstyle/autofix/parser/Configuration.java
```diff
@@ -0,0 +1,41 @@
...
+
+    String[] getPropertyNames();
+
+    String getProperty(String name) throws CheckstyleException;
+
+    Configuration[] getChildren();
+
+    String getName();
+
+    Map<String, String> getMessages();
```
rdiachenko, line 39: remove all not needed, keep only those field that are needed for recipes, name them properly

src/main/resources/META-INF/rewrite/recipes.yml
```diff
@@ -15,3 +15,4 @@ tags:
 recipeList:
   - org.checkstyle.autofix.CheckstyleAutoFix:
       violationReportPath: "target/checkstyle/checkstyle-report.xml"
+      checkstyleConfigurationPath: "https://raw.githubusercontent.com/checkstyle/checkstyle/checkstyle-${checkstyle.version}/config/checkstyle-checks.xml"
```
rdiachenko, line 18: let's get rid of "checkstyle" prefix in name and keep it consistent with "violationReportPath" name as both properties already under CheckstyleXXX recipe

src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
```diff
@@ -49,12 +69,38 @@ public String getViolationReportPath() {
...
     @Override
     public List<Recipe> getRecipeList() {
-
         final List<CheckstyleViolation> violations = CheckstyleReportParser
                 .parse(Path.of(getViolationReportPath()));
 
         return CheckstyleRecipeRegistry.getRecipes(violations);
     }
+
+    private Configuration loadCheckstyleConfiguration() throws CheckstyleException, IOException {
```
rdiachenko, line 88: Let's use same approach as with violation parser. You need to translate `Configuration` object returned by Checkstyle library to our custom simplified `Configuration` object

rdiachenko (changes requested):

src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
Anmol202005: https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/api/Configuration.java

@rdiachenko This interface contains a `public Configuration[] getChildren()` which returns Configurations.

How are we actually looking to map this.
As far as i understand we are looking to create a Model Class Configuration. 
and we have to use it rather than Checkstyle Configuration `INTERFACE`.

the `interface` does not contain a method to return properties as a `MAP` it return `propertyName` as a array.

kindly help out. please clear if i am  missing everything.
if we are looking to make a Model class then it will be a lot of workaround while mapping it.
idk if we want that.

rdiachenko: If we need a map of properties, it makes sense to use it in the new model. The class that will map Checkstyle's configuration into a new model will be responsible for translating `getChildren()` into a map object, if we need this. Make sure we don't miss anything by doing this translation. If we still need a list of configuration children, we can reproduce the structure in the new model

src/main/java/org/checkstyle/autofix/parser/Configuration.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/parser/Configuration.java
Anmol202005: done.

src/main/resources/META-INF/rewrite/recipes.yml
Anmol202005: done.

src/main/java/org/checkstyle/autofix/parser/CheckConfiguration.java
```diff
@@ -0,0 +1,39 @@
...
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package org.checkstyle.autofix.parser;
+
+import java.util.Map;
+
+public final class CheckConfiguration {
+    private final Map<String, String> properties;
+    private final CheckConfiguration[] children;
```
rdiachenko, line 24: Let's use List instead of a raw array as it is more flexible and safer

src/main/java/org/checkstyle/autofix/parser/ConfigurationMapper.java
```diff
@@ -0,0 +1,56 @@
...
+        final String[] propertyNames = config.getPropertyNames();
+        for (String propertyName : propertyNames) {
+            try {
+                final String value = config.getProperty(propertyName);
+                properties.put(propertyName, value);
+
+            }
+            catch (CheckstyleException exception) {
+                throw new IllegalArgumentException("Error getting property "
+                        + propertyName + ": " + exception.getMessage());
```
rdiachenko, line 43: 1. `exception.getMessage()` you are hiding the cause by doing this, keep `exception` as a second parameter
2. `IllegalArgumentException` I think `IllegalStateException` is more appropriate here

src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
```diff
@@ -49,12 +70,39 @@ public String getViolationReportPath() {
...
     public List<Recipe> getRecipeList() {
-
         final List<CheckstyleViolation> violations = CheckstyleReportParser
                 .parse(Path.of(getViolationReportPath()));
 
         return CheckstyleRecipeRegistry.getRecipes(violations);
     }
+
+    private CheckConfiguration loadCheckstyleConfiguration()
+            throws CheckstyleException, IOException {
```
rdiachenko, line 90: 1. Why do we need `throws CheckstyleException, IOException`?
2. This method should be part of the `ConfigurationMapper`. We may want to rename `ConfigurationMapper -> ConfigurationLoader` and make it have a single public method that accepts config path and returns configuration object

src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
```diff
@@ -49,12 +70,39 @@ public String getViolationReportPath() {
         return violationReportPath;
     }
 
+    public String getConfigurationPath() {
+        return configurationPath;
+    }
+
+    public String getPropertiesPath() {
+        return propertiesPath;
+    }
```
rdiachenko, line 79: This is so odd that to load checkstyle config we need two files. @romani is this expected?

src/main/java/org/checkstyle/autofix/parser/CheckConfiguration.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/parser/ConfigurationMapper.java
rdiachenko: not done, you concatinate exception object instead of using it as a second parameter

src/main/java/org/checkstyle/autofix/parser/ConfigurationLoader.java
```diff
@@ -0,0 +1,91 @@
...
+
+        return new CheckConfiguration(properties, List.of(simpleChildren));
+    }
+
+    public static CheckConfiguration loadConfiguration(String checkstyleConfigurationPath,
+                                                       String propFile) {
+        Properties props = new Properties();
+
+        if (propFile == null) {
+            props = System.getProperties();
```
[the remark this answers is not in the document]
Anmol202005: Trying to fetch properties defined locally it seems logical to me. Let me know if I'm misunderstanding: [reference](https://github.com/checkstyle/checkstyle/blob/master/src/main/java/com/puppycrawl/tools/checkstyle/Main.java#L370)

src/main/java/org/checkstyle/autofix/parser/ConfigurationLoader.java
```diff
@@ -0,0 +1,91 @@
...
+import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
+import com.puppycrawl.tools.checkstyle.api.Configuration;
+
+public final class ConfigurationLoader {
+
+    private ConfigurationLoader() {
+        // utility class
+    }
+
+    public static CheckConfiguration mapConfiguration(Configuration config) {
```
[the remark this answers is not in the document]
Anmol202005: done.

src/main/java/org/checkstyle/autofix/parser/ConfigurationMapper.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
timurt: As far as I understand, the properties file is optional in cases where we want to provide different values for different checkstyle configurations

If so, I would support an argument to ignore it for a while, until there is real neediness

Are there any arguments on why we should do it now? @Anmol202005

src/main/java/org/checkstyle/autofix/parser/CheckConfiguration.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public Map<String, String> getProperties() {
+        return properties;
+    }
+
+    public List<CheckConfiguration> getChildren() {
+        return children;
+    }
+
+    public String getProperty(String childName) {
```
timurt, line 52: let's add more getxxx methods

`getIntProperty()`
`getBooleanProperty()`

so that we will hide data type conversion in this class and will avoid having boilerplate code across the codebase 

what do you think?

src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
rdiachenko: @Anmol202005 ping

Anmol202005: we will need to add it at some point. so why not now.

Anmol202005: one more point
our own project requires a property file, cant be ignored i guess !!

rdiachenko: yes, as we don't use it in the header recipe

Anmol202005: am i missing something!!
we specify the header file path in our .properties.

rdiachenko (approved):
ok, let's merge it as it blocks other issues, we will iterate on the logic and improve as part of PRs where we start using it

Issue #41, opened 2025-07-09
Setup integration tests pipeline


rdiachenko:
We need a way to avoid putting manually generated Checkstyle reports and configs while testing recipes. Ideally we should have InputXXX and OutputXXX java files only and a simple test class for a recipe. Thus, we need to setup a pipeline for integraiton test, which will include running Checkstyle, getting violations, passing config and violations to the recipe, running the recipe and verifying the output from the recipe matches expected OuputXXX.

The concept is as follows.

- Add Checkstyle dependency with a `tests` qualifier under the test scope
- Take one recipe and create InputXXX for it following Checkstyle format, which means the InputXXX starts with the check's config
- Check how `AbstractXmlTestSupport` class works. We can create `AbstractRecipeTestSupport` class that will extend `AbstractXmlTestSupport` and have these methods:
```java
abstract getSubpackage(); // subpackage for input files

verify(InputXXX) { // main method that executes pipeline
  // create check based on inline config in InputXXX
  // register XML logger
  // run the check
  // take config and xml report from the check's run and initialise main CheckstyleAutoFix recipe
  // preprocess InputXXX by renaming it to OutputXXX
  // run CheckstyleAutoFix
  // postprocess OutputXXX by removing "// violation" comments
  // verify that obtained OutputXXX matches expected OutputXXX
}
```
- Every recipe test should extend `AbstractRecipeTestSupport` and look like this:
```java
class UpperEllTest extends AbstractRecipeTestSupport {
  String getSubpackage() { return "upperell"; }

  @Test test1() { verify("UpperEllWithXXX"); }

  @Test test2() { verify("UpperEllWithYYY"); }
}
```
- Update all current tests and remove redundant report and config files.

rdiachenko:
From Discord discussion:

> We shouldn't create instances of nested recipes in tests but use only CheckstyleAutoFix for testing. It should create needed nested recipes depending on the violation report and config from the input files.
> Consider this end-to-end test where you have config from the input file, you have violation report from running Checkstyle and you call a single parent CheckstyleAutoFix recipe to fix it and check output.

Anmol202005:
based on: https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/60#issuecomment-3098057560
- Enforce presence of // violation comments in Checkstyle test input files.
- Validate fixed input files by re-running Checkstyle.

[...]

PR #44, opened 2025-07-16
Issue #41: Updated AbstractRecipeTest to remove external report

Anmol202005:
Part of #41 

Updated `AbstractRecipeTest` to ruCheckstyle and generate report directly rather than using an external report.

#### Also Changes made in upperEll recipe:
```java
                if (((J.Literal) targetElement).getValueSource().matches("^[+-].*")) {
                    column++;
                }
```
when the literal has a sign prefix like -10000 or +10000, checkstyle violation point to the first digit (the 1) instead of the sign character. therefore added a check for it.

before:
`result = out.length() - lineBreakIndex - 1;`

after
`result = out.length() - lineBreakIndex;`

We were initially calculating column on  zero-based index. fixed it now. (Checkstyle violations are (1-based))


Anmol202005:
> @rdiachenko Need help here !! not sure why the test fails for window.

fixed it!
The issue was that:
- On Windows: System.lineSeparator() returns \r\n
- Space.format() treats \r and \n as separate characters: 
```java
case '\n':
case '\r':
    if (inSingleLineComment) {
        // Processes single line comments
        comments.add(new TextComment(false, comment.toString(), prefix.toString(), Markers.EMPTY));
        prefix.setLength(0);
        comment.setLength(0);
        prefix.append(c);
    } else if (!inMultiLineComment) {
        prefix.append(c);
    } else {
        comment.append(c);
    }
    break;
```
Not sure how it passes initially!
will look raise a separate PR for this fix.

rdiachenko:
@Anmol202005 please do the following separation:
1. PR that includes fix for UpperEll with corresponding test case. Create a separate issue and fix this under that issue
2. PR that adds `AbstractRecipeTestSupport`, adds checkstyle dependency and implements verify method. Don't tauch `AbstractRecipeTest` and other tests
3. PR that updates `UpperEllTest` to extend from `AbstractRecipeTestSupport`, remove custom reports and config files, and make it work end-to-end for `UpperEllTest` only. If you need anything from `AbstractRecipeTest`, copy it into `AbstractRecipeTestSupport`
4. PR that updates `HeaderTest` similar to the previous point. At this point no tests use `AbstractRecipeTest`, so we can remove it

rdiachenko:
@Anmol202005 please close this PR if we no longer need it

Issue #50, opened 2025-07-19
Implement Header in `CheckstyleAutoFix`


Anmol202005:
Part of this issue:
-  Refactor the Header recipe to use checkConfiguration for simplification.
-  Register the Header recipe in the recipe registry.
-  Accept propertyPath as an input parameter in CheckstyleAutoFix.

rdiachenko:
Resolved by https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/59

PR #51, opened 2025-07-19
Issue #50: Refactored the header recipe to use checkConfiguration

Anmol202005:
Part of: #50 
simplified the implementation by using checkConfiguration in header.

### DIFF
--- a/src/main/java/org/checkstyle/autofix/recipe/Header.java
+++ b/src/main/java/org/checkstyle/autofix/recipe/Header.java
@@ -28,6 +28,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.checkstyle.autofix.parser.CheckConfiguration;
 import org.checkstyle.autofix.parser.CheckstyleViolation;
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Recipe;
@@ -38,9 +39,6 @@ import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.JavaSourceFile;
 import org.openrewrite.java.tree.Space;
 
-import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
-import com.puppycrawl.tools.checkstyle.api.Configuration;
-
 public class Header extends Recipe {
     private static final String HEADER_PROPERTY = "header";
     private static final String HEADER_FILE_PROPERTY = "headerFile";
@@ -48,13 +46,11 @@ public class Header extends Recipe {
     private static final String CHARSET_PROPERTY = "charset";
 
     private final List<CheckstyleViolation> violations;
-    private final Configuration config;
-    private final Charset charset;
+    private final CheckConfiguration config;
 
-    public Header(List<CheckstyleViolation> violations, Configuration config, Charset charset) {
+    public Header(List<CheckstyleViolation> violations, CheckConfiguration config) {
         this.violations = violations;
         this.config = config;
-        this.charset = charset;
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


src/main/java/org/checkstyle/autofix/recipe/Header.java
```diff
@@ -69,60 +65,53 @@ public String getDescription() {
...
-                    charsetToUse = charset;
+                    charsetToUse = Charset.defaultCharset();
                 }
+
                 final String headerFilePath = config.getProperty(HEADER_FILE_PROPERTY);
                 header = Files.readString(Path.of(headerFilePath), charsetToUse);
             }
         }
-        catch (CheckstyleException | IOException exception) {
+        catch (IOException exception) {
```
rdiachenko, line 96: Thsi try-catch wraps too much code that doesn't throw exceptions, please include only those code lines that requre this block

src/main/java/org/checkstyle/autofix/recipe/Header.java
```diff
@@ -69,60 +65,53 @@ public String getDescription() {
...
             else {
                 final Charset charsetToUse;
-                if (hasProperty(config, CHARSET_PROPERTY)) {
+                if (config.hasProperty(CHARSET_PROPERTY)) {
                     charsetToUse = Charset.forName(config.getProperty(CHARSET_PROPERTY));
                 }
+                else if (config.getParent().hasProperty(CHARSET_PROPERTY)) {
+                    charsetToUse = Charset.forName(config
+                            .getParent().getProperty(CHARSET_PROPERTY));
+                }
```
rdiachenko, line 87: I think we should make `getParent()` method private and encapsulate this logic in `hasProperty` and `getProperty`. E.g., `getProperty` internally goes via parents and tryes to find requested property and returns it. This will allow to have simpler contract and easier usage of the config class as the complexity will be hidden.

src/main/java/org/checkstyle/autofix/recipe/Header.java
```diff
@@ -69,60 +65,53 @@ public String getDescription() {
...
                 else {
-                    charsetToUse = charset;
+                    charsetToUse = Charset.defaultCharset();
```
rdiachenko, line 89: Does it make sense to add `getPropertyOrDefault` into config class to handle such cases?

src/main/java/org/checkstyle/autofix/recipe/Header.java
```diff
@@ -69,60 +65,53 @@ public String getDescription() {
...
-                ignoreLinesList = new ArrayList<>();
-            }
-            else {
-                final String ignoreLines = config.getProperty(IGNORE_LINES_PROPERTY);
-                ignoreLinesList = Arrays.stream(ignoreLines.split(","))
-                        .map(String::trim)
-                        .map(Integer::parseInt)
-                        .collect(Collectors.toList());
-            }
+        if (!config.hasProperty(IGNORE_LINES_PROPERTY)) {
```
rdiachenko, line 104: Swap this condition with `else`. E.g., what's easier to read: `if (config.hasProperty(...))` or `if (!config.hasProperty(...))` ?

src/test/java/org/checkstyle/autofix/recipe/HeaderTest.java
```diff
@@ -42,15 +40,14 @@ protected Recipe getRecipe() throws CheckstyleException {
...
-                configPath, new PropertiesExpander(new Properties())
-        );
+        final CheckConfiguration config = ConfigurationLoader.loadConfiguration(configPath);
 
         final List<CheckstyleViolation> violations =
                 CheckstyleReportParser.parse(Path.of(reportPath));
 
-        return new Header(violations,
-                extractCheckConfiguration(config, "Header"), getCharset(config));
+        final CheckConfiguration checkConfig = config.getChild("Header");
```
rdiachenko, line 48: 1. Could you rename `getChild` to `getChildConfig`
2. Using such hardcoded values `"Header"` may be error prone, any type or multiple places and we are in trouble. Let's handle this separately, as an option we can create an enum and put there all checks the we have recipes for. Please create and issue for this

rdiachenko (changes requested):

src/test/java/org/checkstyle/autofix/recipe/HeaderTest.java
Anmol202005: We're already planning to shift to using `CheckstyleAutoFix` directly, so I don't think we'll end up using `CheckEnum`.
Adding it for now to make this mergable, but it's likely to be removed later.

src/main/java/org/checkstyle/autofix/recipe/Header.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/recipe/Header.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/recipe/Header.java
Anmol202005: done.

PR #52, opened 2025-07-19
Issue #50: Accept propertypath as input param

Anmol202005:
part of: #50

### DIFF
--- a/src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
+++ b/src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
@@ -42,6 +42,12 @@ public class CheckstyleAutoFix extends Recipe {
             example = "config/checkstyle.xml")
     private String configurationPath;
 
+    @Option(displayName = "Checkstyle properties file path",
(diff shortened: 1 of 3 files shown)


src/main/java/org/checkstyle/autofix/CheckstyleAutoFix.java
```diff
@@ -42,6 +42,12 @@ public class CheckstyleAutoFix extends Recipe {
             example = "config/checkstyle.xml")
     private String configurationPath;
 
+    @Option(displayName = "Checkstyle properties file path",
+            description = "Path to the file containing the Checkstyle Properties.",
+            example = "config/checkstyle.properties",
+            required = false)
+    private String propertiesPath;
```
rdiachenko, line 49: I assume, this property is not mandatory. Could you check how OpenRewrite suggests to implement optional properties? Here we'll get null if it is not set, but I'm curious if we can mark it with some optional annotation or make it return `Optional<String>`.

rdiachenko (approved):
lgtm

PR #57, opened 2025-07-19
Issue #41: Added AbstractRecipeTestSupport

Anmol202005:
Part of: #41 
Added AbstractRecipeTestSupport with `verify` method and added checkstyle dependency.


rdiachenko (approved):
Most probably we'll need to update it to create parent recipe only -- CheckstyleAutoFix -- and rely on it to handle nested recipe creation and execution. We will also need to update verify with postprocessing mentioned in the issue. But this is a good start, good job @Anmol202005 !

PR #59, opened 2025-07-21
Issue #50: Register the Header in recipe registry

Anmol202005:
Fixes: #50 
Implemented Header recipe in recipe registry


src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
```diff
@@ -47,28 +55,55 @@ private CheckstyleRecipeRegistry() {
...
-                    .substring(checkName.lastIndexOf('.') + 1);
-            final List<CheckstyleViolation> checkViolations = entry.getValue();
+    private static Recipe createRecipe(Map.Entry<String, List<CheckstyleViolation>> entry,
+                                       CheckConfiguration config) {
+        final String simpleCheckName = normalizeCheckName(entry.getKey());
+        final List<CheckstyleViolation> violations = entry.getValue();
 
-            final Function<List<CheckstyleViolation>, Recipe> recipeFactory =
+        Recipe recipe = null;
+        final BiFunction<List<CheckstyleViolation>, CheckConfiguration, Recipe> configFactory =
```
timurt, line 78: I know that some people prefer explicit `if`, but I am more in favor of chaining :) 

ex. 

what do you think about this modification?

```
        return Optional.ofNullable(RECIPE_MAP_WITH_CONFIG.get(simpleCheckName))
                .map(factory -> factory.apply(violations, extractCheckConfiguration(config, simpleCheckName)))
                .orElseGet(() -> Optional.ofNullable(RECIPE_MAP.get(simpleCheckName))
                        .map(factory -> factory.apply(violations))
                        .orElse(null));
```

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
```diff
@@ -47,28 +55,55 @@ private CheckstyleRecipeRegistry() {
...
+    }
+
+    private static CheckConfiguration extractCheckConfiguration(CheckConfiguration config,
+                                                                String checkName) {
+        return config.getChildConfig(checkName);
+    }
+
+    private static String normalizeCheckName(String checkName) {
+        String normalizedCheckName = checkName.substring(checkName.lastIndexOf('.') + 1);
+        final int checkLength = 5;
```
timurt, line 102: minor: I think this variable is redundant, I see no problem to do just `.substring(0, normalizedCheckName.length() - 5);`

timurt (commented):
Minor comments from my side

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
Anmol202005: `magicNumber check` fails the build.
https://checkstyle.sourceforge.io/checks/coding/magicnumber.html

src/test/java/org/checkstyle/autofix/CheckstyleRecipeRegistryTest.java
```diff
@@ -22,10 +22,20 @@
...
+import org.checkstyle.autofix.parser.ConfigurationLoader;
 import org.junit.jupiter.api.Test;
 import org.openrewrite.Recipe;
 
-public class CheckstyleRecipeRegistryTest {
+import com.puppycrawl.tools.checkstyle.AbstractXmlTestSupport;
+import com.puppycrawl.tools.checkstyle.DefaultConfiguration;
+import com.puppycrawl.tools.checkstyle.checks.header.HeaderCheck;
+
+public class CheckstyleRecipeRegistryTest extends AbstractXmlTestSupport {
```
rdiachenko, line 33: I assume, we won't need this test after integration pipeline is complete. So when calling CheckstyleAutoFix it will implicitly cover CheckstyleRecipeRegistry

src/main/java/org/checkstyle/autofix/parser/CheckConfiguration.java
```diff
@@ -96,11 +100,15 @@ public int[] getIntArray(String propertyName) {
 
     public CheckConfiguration getChildConfig(String childName) {
         CheckConfiguration result = null;
-        for (CheckConfiguration child : children) {
+        for (CheckConfiguration child : getChildren()) {
             if (childName.equals(child.getName())) {
                 result = child;
                 break;
             }
+            result = child.getChildConfig(childName);
```
rdiachenko, line 108: could you rewrite this method without using recursion, and using a simple while loop for better readability?

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
rdiachenko: Why do we need `normalizeCheckName` at all? Can we avoid it?

timurt (approved):
LGTM from my side

Thank you!

src/main/java/org/checkstyle/autofix/parser/CheckConfiguration.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
```diff
@@ -47,28 +56,49 @@ private CheckstyleRecipeRegistry() {
...
-        final List<Recipe> recipes = new ArrayList<>();
+        return Optional.ofNullable(RECIPE_MAP_WITH_CONFIG.get(simpleCheckName))
+                .map(factory -> {
+                    return factory.apply(violations,
+                            extractCheckConfiguration(config, simpleCheckName));
+                }).orElseGet(() -> {
+                    return Optional.ofNullable(RECIPE_MAP.get(simpleCheckName))
+                            .map(factory -> factory.apply(violations))
+                            .orElse(null);
+                });
```
rdiachenko, line 87: could you use just simple if-else to avoid this ofNullable and orElse complexity?

rdiachenko (changes requested):

src/main/java/org/checkstyle/autofix/parser/CheckConfiguration.java
rdiachenko: you can get rid of `index` by using iterator
```java
for (Iterator<..> iter = childrenList.iterator(); iter.hasNext();) {
...
iter.next();
...
}
```

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/CheckstyleCheck.java
```diff
@@ -0,0 +1,43 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package org.checkstyle.autofix;
+
+public enum CheckstyleCheck {
+    UPPERELL("com.puppycrawl.tools.checkstyle.checks.UpperEllCheck"),
+    HEADER("header");
```
rdiachenko, line 22: This is veird, can you confirm Checkstyle doesn't provide package prefixed path for it? I expect it to be similar to UpperEllCheck. If this is not the case, please reaise a question in the main checkstyle channel

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
rdiachenko: We should return `CheckstyleCheck` object and callers should work with it. Additionally, please read a code snipped again, looks like you missed what is written in comments https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/59#discussion_r2231870862

Anmol202005: > We should return `CheckstyleCheck` object and callers should work with it. Additionally, please read a code snipped again, looks like you missed what is written in comments [#59 (comment)](https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/59#discussion_r2231870862)

@rdiachenko throwing an exception does not makes sense here.
am i missing something ?
we need to fix those checks that are present in the `Enum` but should not fail if a check fix is not present.

src/main/java/org/checkstyle/autofix/CheckstyleCheck.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
rdiachenko: ok, I see what you mean, the report may contain violations from unsupported checks and we don't want to fail, yes that makes sense. But in your logic you'll still fail with NPE here: `final String checkstyleCheck = CheckstyleCheck.fromSource(entry.getKey()).name();`

Please do the following:
1. Update `fromSource` to return `Optional<CheckstyleCheck>` and update the caller side accordingly
2. You missed this point from my initial comment: `Remove normalization method, and use EnumMap where the key is CheckstyleCheck.` and you still use hardcoded string values:
```java
RECIPE_MAP.put("UPPERELL", UpperEll::new);
RECIPE_MAP_WITH_CONFIG.put("HEADER", Header::new);
```
Read about `EnumMap` and update the logic as requested

Anmol202005: done.

rdiachenko (approved):
It is not ideal:
- `createRecipe` could return optional instead of null
- `if (check.isPresent())` could be replaced with more readable stream api like `check.ifPresent`

But we already spent too much time on this PR, and it blocks further progress. I'm ok to merge it

PR #62, opened 2025-07-22
Issue #41: Updated headerTest to use AbstractRecipeTestSupport

Anmol202005:
Part of: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/41
Updated `HeaderTest` to extend `AbstractRecipeTestSupport`.


rdiachenko:
@Anmol202005 

Let's step back. Why do we need ` + System.lineSeparator()` at all? I think we should rely on what is in the header file and just concatinate it.

Additionally, could you resolve conflicts please

Anmol202005:
@rdiachenko Done.

src/test/java/org/checkstyle/autofix/recipe/HeaderTest.java
```diff
@@ -17,53 +17,57 @@
...
 
     @Test
-    void headerCommentTest() throws IOException, CheckstyleException {
-        testRecipe("header", "HeaderComments");
+    void headerCommentTest() throws Exception {
+        final DefaultConfiguration checkConfig = createModuleConfig(HeaderCheck.class);
+        final String headerPath = "src/test/resources/org/checkstyle/autofix/"
+                + "recipe/header/header.txt";
+        checkConfig.addProperty("headerFile", headerPath);
+        checkConfig.addProperty("ignoreLines", "3");
```
rdiachenko, line 60: Let's single out it into a private method and reuse it in all tests as this config is the same in all cases. Additionally, could you confirm we can't configure it via InputXXX as for other cases?

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
```diff
@@ -46,29 +48,31 @@ public abstract class AbstractRecipeTestSupport extends AbstractXmlTestSupport
...
-        final Configuration config = getCheckConfigurations(inputPath);
         final List<CheckstyleViolation> violations = runCheckstyle(inputPath, config);
 
         final String beforeCode = readFile(getPath(inputPath));
         final String expectedAfterCode = readFile(getPath(outputPath));
-
-        final Recipe mainRecipe = createRecipe(violations);
-
+        final CheckConfiguration checkConfig = ConfigurationLoader.mapConfiguration(config);
+        final Recipe mainRecipe = createRecipe(violations, checkConfig);
```
rdiachenko, line 75: Could you create an issue to explore how we can avoid creating nested recipes but instead use a parent CheckstyleAutoFix recipe that handles configuration and nested recipes creation itself, so we can simplify testing logic

src/test/java/org/checkstyle/autofix/recipe/HeaderTest.java
rdiachenko: We probably should rethink the approach taken in the Header recipe. This is a good example when we may have some section commented out as the file header. So we don't want to overwrite it but instead preppend the license header as the first block of comments. Worth creating an issue for this I think

Anmol202005: @rdiachenko apart for the recipe, are we sure that the header check will work fine with config comment(checkstyle input file dont use a config in input file for header).

rdiachenko: > checkstyle input file dont use a config in input file for header

this is what I asked you to confirm :) `could you confirm we can't configure it via InputXXX as for other cases?`

ok, let's keep it as it is now, just single out commont logic into a separate method and reuse it in all 3 tests

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
Anmol202005: Issue created: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/67

src/test/java/org/checkstyle/autofix/recipe/HeaderTest.java
Anmol202005: > We probably should rethink the approach taken in the Header recipe. This is a good example when we may have some section commented out as the file header. So we don't want to overwrite it but instead preppend the license header as the first block of comments. Worth creating an issue for this I think

done. https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/68

rdiachenko (approved):
lgtm

Issue #69, opened 2025-07-25
Bug: Fix suffix handling in `RemoveViolationComments`


Anmol202005:
When `RemoveViolationComments` removes a violation comment, it incorrectly adds the removed comment's suffix to the last remaining comment instead of the comment that immediately precedes the removed comment.

Current behavior:
The suffix is added to the last comment in the filtered list, which may not be the actual previous comment.

Expected behavior:
The suffix should be added to the previous comment (the one that appears right before the violation comment in the original list).

#### InputCode
```java
public class Test {
    // comment1
    // violation some text

    // comment2
    int x = 1;
}
```

#### Current Implementation
```java
public class Test {
    // comment1
    // comment2

    int x = 1;
}
```

#### Expected Implementation
```java
public class Test {
    // comment1

    // comment2
    int x = 1;
}
```

Anmol202005:
presequisite for #64

Anmol202005:
@rdiachenko good to close ?

rdiachenko:
Resolved by https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/70

PR #70, opened 2025-07-25
Issue #69: resolved removeViolationComments Bug

Anmol202005:
Fixes: #69

### DIFF
--- a/src/test/java/org/checkstyle/autofix/RemoveViolationComments.java
+++ b/src/test/java/org/checkstyle/autofix/RemoveViolationComments.java
@@ -17,9 +17,8 @@
 
 package org.checkstyle.autofix;
 
+import java.util.ArrayList;
 import java.util.List;
-import java.util.Objects;
-import java.util.stream.Collectors;
 
 import org.openrewrite.ExecutionContext;
 import org.openrewrite.Recipe;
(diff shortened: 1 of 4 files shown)


rdiachenko (changes requested):
ok, this is the third time you fix it, I'm now asking to cover it by tests, all possible edge cases please

src/test/java/org/checkstyle/autofix/RemoveViolationComments.java
```diff
@@ -49,21 +48,31 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
-                            if (textComment.getText().startsWith("violation")) {
-                                suffixAccumulator.append(textComment.getSuffix());
-                                result = null;
-                            }
+            for (Comment comment : space.getComments()) {
+                Comment result = comment;
+                if (!comment.isMultiline() && comment instanceof TextComment) {
+                    final TextComment textComment = (TextComment) comment;
+                    if (textComment.getText().matches("\\s*(\\d+\\s*)?violation.*")) {
+                        if (!filteredComments.isEmpty()) {
```
rdiachenko, line 59: swap condition, follow this pattern for better code readability: `if (positive condition) {...} else {...}`
This is not the first time I see this.
What's easier to read? `if (filteredComments.isEmpty())` or `if (!filteredComments.isEmpty())`

src/test/java/org/checkstyle/autofix/RemoveViolationComments.java
```diff
@@ -49,21 +48,31 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
+                            filteredComments.set(lastIndex,
+                                    previousComment.withSuffix(textComment.getSuffix()));
                         }
-                        return result; })
-                    .filter(Objects::nonNull)
-                    .collect(Collectors.toList());
+                        else {
+                            suffixAccumulator = textComment.getSuffix();
+                        }
+                        result = null;
```
rdiachenko, line 68: instead of setting it null and later checking for null you may extend the outer if condition with `else { filteredComments.add(result); }`

src/test/java/org/checkstyle/autofix/RemoveViolationComments.java
Anmol202005: yep, done
will be careful with this.

rdiachenko (approved):
not ideal, logic is too complex and not modular, but this class is used for tests purposes only and we already spent too much time on it, ok to merge

PR #76, opened 2025-08-05
Issue #41: added CheckstyleAutoFix as Test Recipe

Anmol202005:
Fixes: #41 

Implemented CheckstyleAutoFix in Test pipeline.

### DIFF
--- a/src/main/java/org/checkstyle/autofix/CheckstyleCheck.java
+++ b/src/main/java/org/checkstyle/autofix/CheckstyleCheck.java
@@ -22,7 +22,8 @@ import java.util.Optional;
 
 public enum CheckstyleCheck {
     UPPERELL("com.puppycrawl.tools.checkstyle.checks.UpperEllCheck"),
-    HEADER("com.puppycrawl.tools.checkstyle.checks.header.HeaderCheck");
+    HEADER("com.puppycrawl.tools.checkstyle.checks.header.HeaderCheck"),
+    FINALLOCALVARIABLE("com.puppycrawl.tools.checkstyle.checks.coding.FinalLocalVariableCheck");
 
     private final String id;
 
--- a/src/main/java/org/checkstyle/autofix/parser/ConfigurationLoader.java
+++ b/src/main/java/org/checkstyle/autofix/parser/ConfigurationLoader.java
@@ -25,6 +25,8 @@ import java.util.Locale;
 import java.util.Map;
 import java.util.Properties;
 
+import org.checkstyle.autofix.CheckstyleCheck;
+
 import com.puppycrawl.tools.checkstyle.PropertiesExpander;
 import com.puppycrawl.tools.checkstyle.api.CheckstyleException;
 import com.puppycrawl.tools.checkstyle.api.Configuration;
@@ -56,8 +58,12 @@ public final class ConfigurationLoader {
         for (int index = 0; index < checkstyleChildren.length; index++) {
             simpleChildren[index] = mapConfiguration(checkstyleChildren[index]);
         }
+        String moduleName = config.getName().toUpperCase(Locale.ROOT);
+        if (CheckstyleCheck.fromSource(config.getName()).isPresent()) {
+            moduleName = CheckstyleCheck.fromSource(config.getName()).get().name();
+        }
 
-        return new CheckConfiguration(config.getName().toUpperCase(Locale.ROOT),
+        return new CheckConfiguration(moduleName,
                 properties, List.of(simpleChildren));
     }
 
--- a/src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
+++ b/src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
@@ -20,13 +20,17 @@ package org.checkstyle.autofix.recipe;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.openrewrite.java.Assertions.java;
 
+import java.io.BufferedWriter;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
+import java.io.FileWriter;
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Collections;
 import java.util.List;
 
+import org.checkstyle.autofix.CheckstyleAutoFix;
 import org.checkstyle.autofix.InputClassRenamer;
 import org.checkstyle.autofix.RemoveViolationComments;
 import org.checkstyle.autofix.parser.CheckConfiguration;
@@ -48,9 +52,6 @@ public abstract class AbstractRecipeTestSupport extends AbstractXmlTestSupport
 
     protected abstract String getSubpackage();
 
-    protected abstract Recipe createRecipe(List<CheckstyleViolation> violations,
-                                           CheckConfiguration configuration);
-
     @Override
     protected String getPackageLocation() {
         return "org/checkstyle/autofix/recipe/" + getSubpackage();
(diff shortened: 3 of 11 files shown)


src/main/java/org/checkstyle/autofix/parser/ConfigurationLoader.java
```diff
@@ -56,8 +58,12 @@ public static CheckConfiguration mapConfiguration(Configuration config) {
         for (int index = 0; index < checkstyleChildren.length; index++) {
             simpleChildren[index] = mapConfiguration(checkstyleChildren[index]);
         }
+        String moduleName = config.getName().toUpperCase(Locale.ROOT);
+        if (CheckstyleCheck.fromSource(config.getName()).isPresent()) {
+            moduleName = CheckstyleCheck.fromSource(config.getName()).get().name();
+        }
```
rdiachenko, line 64: Why do we still use String names, we should move to `CheckstyleCheck` enum usage

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
```diff
@@ -63,38 +64,58 @@ protected void verify(String testCaseName) throws Exception {
...
         final String expectedAfterCode = readFile(getPath(outputPath));
         final CheckConfiguration checkConfig = ConfigurationLoader.mapConfiguration(config);
-        final Recipe mainRecipe = createRecipe(violations, checkConfig);
-        testRecipe(beforeCode, expectedAfterCode,
-                getPath(inputPath), new InputClassRenamer(),
-                mainRecipe, new RemoveViolationComments());
+        final Path configPath = Files.createTempFile("checkstyle-config", ".xml");
+
+        try {
+            writeConfigurationToXml(checkConfig, configPath.toString());
```
rdiachenko, line 78: Let's single this out into a separate method that receives `Configuration config` and returns a `Path` to it

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
```diff
@@ -63,38 +64,58 @@ protected void verify(String testCaseName) throws Exception {
...
+            final Recipe mainRecipe = new CheckstyleAutoFix(violations.toString(),
+                    configPath.toString());
```
rdiachenko, line 80: you should probably use a different method, `toString()` on Path may not be reliable. Please check alternatives

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
```diff
@@ -63,38 +64,58 @@ protected void verify(String testCaseName) throws Exception {
         final Configuration config = getCheckConfigurations(inputPath);
         verifyOutputFile(outputPath, config);
 
-        final List<CheckstyleViolation> violations = runCheckstyle(inputPath, config);
+        final Path violations = runCheckstyle(inputPath, config);
```
rdiachenko, line 67: Reanem to violationReportPath

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
```diff
@@ -105,34 +126,46 @@ private List<CheckstyleViolation> runCheckstyle(String inputPath,
...
-                        .append(", col: ").append(violation.getColumn())
-                        .append(", message: ").append(violation.getMessage())
-                        .append("\n");
-            }
+        for (String propName : config.getPropertyNames()) {
+            writer.write("  <property name=\"" + propName + "\" value=\""
+                    + config.getProperty(propName).replace("&", "&amp;")
+                    .replace("<", "&lt;").replace(">", "&gt;")
+                    .replace("\"", "&quot;").replace("'", "&apos;")
+                    + "\"/>\n");
```
rdiachenko, line 154: why do we need this?

src/main/java/org/checkstyle/autofix/parser/ConfigurationLoader.java
Anmol202005: Config contains both fixable and non-fixable checks.
The above logic is used to store fixable check names in the Checkstyle configuration by converting fully qualified class names to simple class names.
Am i missing something?

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
Anmol202005: yep removed, will add only if needed.

src/main/java/org/checkstyle/autofix/CheckstyleCheck.java
```diff
@@ -22,7 +22,8 @@
 
 public enum CheckstyleCheck {
     UPPERELL("com.puppycrawl.tools.checkstyle.checks.UpperEllCheck"),
-    HEADER("com.puppycrawl.tools.checkstyle.checks.header.HeaderCheck");
+    HEADER("com.puppycrawl.tools.checkstyle.checks.header.HeaderCheck"),
+    FINALLOCALVARIABLE("com.puppycrawl.tools.checkstyle.checks.coding.FinalLocalVariableCheck");
```
rdiachenko, line 26: rename it to `FINAL_LOCAL_VARIABLE`
we should do the same for `UPPER_ELL`

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
```diff
@@ -105,34 +123,50 @@ private List<CheckstyleViolation> runCheckstyle(String inputPath,
...
+        final Path configPath = Files.createTempFile("checkstyle-config", ".xml");
+        writeConfigurationToXml(checkConfig, configPath.toString());
+        return configPath;
+    }
 
-        final List<CheckstyleViolation> violations = runCheckstyle(outputPath, config);
-        if (!violations.isEmpty()) {
-            final StringBuilder violationMessage =
-                    new StringBuilder("Checkstyle violations found in the output file:\n");
+    public static void writeConfigurationToXml(CheckConfiguration config, String filePath)
```
rdiachenko, line 138: We have access to input file, why can't we extract the config part and put it into a file?

src/main/java/org/checkstyle/autofix/CheckstyleCheck.java
Anmol202005: IMO we can work around here,
we can convert module name to fullyQualifiedClassName and then if they are present in CheckstyleCheck enum then we can replace the module name with the Enum value. 

this way we can use Upper_Ell.

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
rdiachenko: have you tried to add it? what issues did you face with?

src/main/java/org/checkstyle/autofix/parser/ConfigurationLoader.java
rdiachenko: We should try to move to a single approach. The problem is that `CheckstyleViolation.source` and `CheckConfiguration.name` should match. Please create an issue to explore using `CheckstyleCheck` as a type for both fields to get rid of custom string comparions. This should probably help to simplify a tree structure of configuration, because in reality we just need a single config per check. I'm ok to merge current code to unblock the testing. Reply with the link to the issue

Anmol202005: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/78

Issue #48, opened 2025-07-17
Implement recipe for RedundantImport


rdiachenko:
Check: https://checkstyle.sourceforge.io/checks/imports/redundantimport.html#RedundantImport

Check core OpenRewrite repository and reuse existing recipes if present.

Update README with the recipe link and notes about coverage if needed.

rdiachenko:
@Anmol202005 please update README if it's not done yet before we close it

PR #77, opened 2025-08-07
Issue: #48: Created new Recipe for Redundant Import

Anmol202005:
Fixes: #48

### DIFF
--- a/src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
+++ b/src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
@@ -1,0 +1,126 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle-openrewrite-recipes: Automatically fix Checkstyle violations with OpenRewrite.
+// Copyright (C) 2025 The Checkstyle OpenRewrite Recipes Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package org.checkstyle.autofix.recipe;
+
+import java.nio.file.Path;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.checkstyle.autofix.PositionHelper;
+import org.checkstyle.autofix.parser.CheckstyleViolation;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+
+public class RedundantImport extends Recipe {
+
+    private static final String JAVA_LANG_PREFIX = "java.lang.";
+
+    private final List<CheckstyleViolation> violations;
+
+    public RedundantImport(List<CheckstyleViolation> violations) {
+        this.violations = violations;
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "Remove redundant imports";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Remove duplicate imports, java.lang imports, and same package imports";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new RemoveRedundantImportsVisitor();
+    }
+
+    private final class RemoveRedundantImportsVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        private Path sourcePath;
+
+        @Override
+        public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu, ExecutionContext ctx) {
+
+            this.sourcePath = cu.getSourcePath().toAbsolutePath();
+
+            final Set<String> seenImports = new HashSet<>();
+            final String currentPackage = getCurrentPackage(cu);
+
+            return cu.withImports(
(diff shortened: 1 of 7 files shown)


src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
```diff
@@ -0,0 +1,123 @@
...
+            if (seenImports.contains(importName)) {
+                isRedundant = true;
+            }
+            else {
+                seenImports.add(importName);
+
+                if (importName.startsWith("java.lang.") && !importStmt.isStatic()) {
+                    isRedundant = true;
+                }
+                else if (currentPackage != null && importName.startsWith(currentPackage + ".")) {
```
timurt, line 92: based on the description from https://checkstyle.sourceforge.io/checks/imports/redundantimport.html#Description
> The class non-statically imported is from the same package as the current package. 

I reckon you also want to add `!importStmt.isStatic()` here, static imports from the same package should be allowed, as far as I understand 

@rdiachenko  correct me if I am wrong

src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
```diff
@@ -0,0 +1,123 @@
...
+
+            boolean isRedundant = false;
+
+            if (seenImports.contains(importName)) {
+                isRedundant = true;
+            }
+            else {
+                seenImports.add(importName);
+
+                if (importName.startsWith("java.lang.") && !importStmt.isStatic()) {
```
timurt, line 89: Let's move magic string "java.lang." to a constant field

timurt, line 89: technically `!importStmt.isStatic()` is faster than `importName.startsWith("java.lang.")`, You can switch the two conditions

src/test/resources/org/checkstyle/autofix/recipe/redundantimport/complexcase/InputComplexCase.java
```diff
@@ -0,0 +1,43 @@
...
+  <module name="TreeWalker">
+    <module name="com.puppycrawl.tools.checkstyle.checks.imports.RedundantImportCheck"/>
+  </module>
+</module>
+
+*/
+
+package org.checkstyle.autofix.recipe.redundantimport.complexcase;
+
+import org.checkstyle.autofix.recipe.redundantimport.complexcase.*; // violation 'Redundant import from the same package'
```
timurt, line 12: Can you also please add same package static import

src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
```diff
@@ -0,0 +1,123 @@
...
+                        .getExpression().printTrimmed(getCursor());
+            }
+            return currentPackage;
+        }
+
+        private boolean isAtViolationLocation(J.Import literal) {
+            final J.CompilationUnit cursor = getCursor().firstEnclosing(J.CompilationUnit.class);
+
+            final int line = PositionHelper.computeLinePosition(cursor, literal, getCursor());
+            final int column = PositionHelper.computeColumnPosition(cursor, literal, getCursor());
```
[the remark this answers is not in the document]
Anmol202005: Yes, Column matters. 
for example here:
```java
import java.util.List;
    import java.util.List; // violation 'Duplicate import'
```
violation is at 2:4

src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
Anmol202005: yep. done.
also added a test example.

src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
Anmol202005: done.

src/test/resources/org/checkstyle/autofix/recipe/redundantimport/complexcase/InputComplexCase.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
```diff
@@ -0,0 +1,126 @@
...
+                            .filter(importStmt -> {
+                                return !isRedundant(importStmt, seenImports,
+                                    currentPackage) || !isAtViolationLocation(importStmt); })
+                            .toList()
+            );
+        }
+
+        private boolean isRedundant(J.Import importStmt,
+                                    Set<String> seenImports, String currentPackage) {
+            final String importName = importStmt.getQualid().toString();
```
timurt, line 81: minor, just for discussion, I know that in checkstyle we usually stick to single return statement 

But wdyt the following code structure? Does it make the code more readable 

```
private boolean isRedundant(J.Import importStmt, Set<String> seenImports, String currentPackage) {
    final String importName = importStmt.getQualid().toString();
    
    if (seenImports.contains(importName)) {
        return true;
    }
    
    seenImports.add(importName);
    
    if (importStmt.isStatic()) {
        return false;
    }
    
    if (importName.startsWith(JAVA_LANG_PREFIX)) {
        return true;
    }
    
    if (currentPackage != null && importName.startsWith(currentPackage + ".")) {
        return true;
    }
    
    return false;
}
```

timurt (approved):
LGTM, just added one comment to discuss

src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
timurt: Thank you!

src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
Anmol202005: IMO we should stick with single return.

rdiachenko (approved):
lgtm

src/main/java/org/checkstyle/autofix/recipe/RedundantImport.java
rdiachenko: pros and cons for both approaches, I don't have preference here. Multiple returns makes it easer to read, on the other hand harder to debug, especially when there are lots of them as in the case above

Issue #85, opened 2025-08-08
Bug: Position Helper Ignores the Prefix Comments in the J.Element


Anmol202005:
Current Implementation:
```java
if (targetElement.isScope(getContext().getCursor().getValue())) {
                            super.append(targetElement.getPrefix().getWhitespace());
                            throw new CancellationException();
}
```
Only the Whitespace before the element is appended.

Therefore incorrect position is calculated when the J.element is preceded by comments.
for example:
```java
import java.lang.*; // violation 'Redundant import from the java.lang package'
import java.lang.String; // violation 'Redundant import from the java.lang package'
```
Here the position of `import java.lang.String` is calculated as 1:20
but it should be 2:1

PR #86, opened 2025-08-08
Issue #85: resolved position helper bug

Anmol202005:
Fixes: #85


rdiachenko (commented):
please add a test that fails before applying this fix, and is passed after applying it

Issue #8, opened 2025-06-07
Autofix strategy: Batch fix all modules or fix individual violations


Anmol202005:
We need to decide on the approach for applying fixes generated by the autofix tool after checkstyle identifies violations.

**Two approaches under consideration:**

1. **Fix all modules** - Apply fixes across all modules in the project
2. **Fix by violation** - Apply fixes to specific violation positions

rdiachenko:
**Option 1: Violation-Driven Recipe Execution**

The idea is to use the Checkstyle library as a dependency, run it to get violations, then execute targeted OpenRewrite recipes only on the specific code locations that Checkstyle flagged. Here's how the flow may work:

```
Input: Checkstyle config + Source files
  ↓
1. Run Checkstyle analysis → Get violations with line/column info
  ↓  
2. Map violations to recipes → Group by check type (UpperEll, etc.)
  ↓
3. Execute targeted recipes → Only fix code at violation locations
  ↓
Output: Fixed source files
```

```java
public class ViolationTargetedRecipeRunner {
    public void executeTargetedFixes(Path checkstyleConfig, List<Path> sourceFiles) {
        // 1. Run Checkstyle with suppressions to get actual violations
        List<CheckstyleViolation> violations = runCheckstyleWithSuppressions(checkstyleConfig, sourceFiles);
        
        // 2. Group violations by recipe type
        Map<Class<? extends Recipe>, List<ViolationLocation>> recipeViolations = 
            mapViolationsToRecipes(violations);
        
        // 3. Execute recipes with violation-specific visitors
        recipeViolations.forEach((recipeClass, locations) -> {
            Recipe recipe = createViolationAwareRecipe(recipeClass, locations);
            executeRecipe(recipe, sourceFiles);
        });
    }
}
```

Maven integration may be something like this:

```xml
<!-- Single plugin that runs Checkstyle internally and applies fixes -->
<plugin>
    <groupId>org.openrewrite.maven</groupId>
    <artifactId>rewrite-maven-plugin</artifactId>
    <configuration>
        <activeRecipes>
            <recipe>org.checkstyle.CheckstyleAutoFix</recipe>
        </activeRecipes>
        <checkstyleConfig>config/checkstyle.xml</checkstyleConfig>
    </configuration>
    <executions>
        <execution>
            <goals>
                <goal>run</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

`mvn rewrite:run`

[...]

PR #87, opened 2025-08-09
Pull #87: Apply OpenRewrite recipe best practices

timtebeek:
Hi! Neat to see you're developing recipes to help Checkstyle users fix their issues; figured I'd help out by applying some (customized) best practices to your recipes based on what we've found to work well for our own recipes. Hope that's appreciated, and curious to hear about your plans here! I imagine there's quite a bit of overlap between your plans vs what we already have in rewrite-static-analysis, so wondering what the best way forward would be to collaborate, if at all.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,11 +20,10 @@
         <maven.compiler.plugin>3.11.0</maven.compiler.plugin>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <rewrite.version>8.54.0</rewrite.version>
-        <rewrite.maven.plugin>6.11.0</rewrite.maven.plugin>
-        <recipe.bom.version>3.9.0</recipe.bom.version>
-        <junit.version>5.13.0</junit.version>
-        <assertj.version>3.24.2</assertj.version>
+        <rewrite.maven.plugin>6.16.0</rewrite.maven.plugin>
+        <recipe.bom.version>3.13.0</recipe.bom.version>
+        <junit.version>5.13.4</junit.version>
+        <assertj.version>3.27.4</assertj.version>
 
         <!-- Checkstyle properties -->
         <maven.checkstyle.plugin.version>3.6.0</maven.checkstyle.plugin.version>
@@ -56,12 +55,10 @@
         <dependency>
             <groupId>org.openrewrite</groupId>
             <artifactId>rewrite-java</artifactId>
-            <version>${rewrite.version}</version>
         </dependency>
         <dependency>
             <groupId>org.openrewrite</groupId>
             <artifactId>rewrite-java-17</artifactId>
-            <version>${rewrite.version}</version>
         </dependency>
 
         <!-- JGit for generating diffs without external Git dependency -->
@@ -88,7 +85,6 @@
         <dependency>
             <groupId>org.openrewrite</groupId>
             <artifactId>rewrite-test</artifactId>
-            <version>${rewrite.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
@@ -182,10 +178,19 @@
                     <exclusions>
                         <exclusion>src/test/resources/**</exclusion>
                     </exclusions>
-                    <recipeArtifactCoordinates>
-                        <coordinate>org.checkstyle.autofix:checkstyle-openrewrite-recipes:1.0.0</coordinate>
-                    </recipeArtifactCoordinates>
                 </configuration>
+                <dependencies>
+                    <dependency>
+                        <groupId>org.checkstyle.autofix</groupId>
+                        <artifactId>checkstyle-openrewrite-recipes</artifactId>
+                        <version>${project.version}</version>
+                    </dependency>
+                    <dependency>
+                        <groupId>org.openrewrite.recipe</groupId>
+                        <artifactId>rewrite-rewrite</artifactId>
+                        <version>RELEASE</version>
+                    </dependency>
+                </dependencies>
                 <executions>
                     <execution>
                         <id>checkstyle-autofix</id>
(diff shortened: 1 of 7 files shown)


rdiachenko (commented):
@timtebeek thanks a lot for the PR!

> curious to hear about your plans here! I imagine there's quite a bit of overlap between your plans vs what we already have in rewrite-static-analysis, so wondering what the best way forward would be to collaborate, if at all.

We started this project as part of GSoC this summer, so it's still at a very early stage. There are indeed overlaps with some existing recipes, and I expect we might reuse them at some point.

The main problem we're trying to solve is having a more granular suppression model. OpenRewrite doesn't allow suppression by line and column within a single file. This makes its practical usage limited. Our approach is to fix only the code that actually violates a rule.

Here's a simple example. Let's say we enable Checkstyle's `UpperEll` check but suppress all constants:
```java
class Main {
  final long foo = 8l;  // no violation, suppressed
  long bar = 8l;        // violation 'Should use uppercase 'L'
}
```

When we run the recipe for `UpperEll`, we expect this ouput:
```java
class Main {
  final long foo = 8l;  // no changes
  long bar = 8L;        // 'l' replaced with 'L'
}
```

We disscussed few approaches to handle this, which you may check here for more context: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/8#issuecomment-2954310499.

The main idea we're evaluating in this project is report-based recipe execution. Checkstyle (and similar tools) generate reports containing violations and their exact  locations in the code. We can use this information to automatically fix only the parts of the code that actually violate the rules.

I'm curious to hear your thoughts on this. Are there any plans in OpenRewrite to extend the suppression model?

cc @romani

rdiachenko:
> Optionally we have some flows documented to run recipes on PRs and create code suggestions; that might help you here as well.

Could you share a link to repo where is it done, or some relevant docs we could check?

timtebeek:
> > Optionally we have some flows documented to run recipes on PRs and create code suggestions; that might help you here as well.
> 
> Could you share a link to repo where is it done, or some relevant docs we could check?

Sure! That's [documented here](https://www.moderne.io/blog/stop-breaking-ci-annotate-prs-with-openrewrite-recipe-fixes-as-quality-gate) and for instance implemented with these two workflows:
https://github.com/apache/camel-upgrade-recipes/blob/main/.github/workflows/receive-pr.yml
https://github.com/apache/camel-upgrade-recipes/blob/main/.github/workflows/comment-pr.yml

src/test/java/org/checkstyle/autofix/recipe/RemoveViolationCommentsTest.java
```diff
@@ -28,7 +28,7 @@
 
 import com.puppycrawl.tools.checkstyle.AbstractPathTestSupport;
 
-public class RemoveViolationCommentsTest extends AbstractPathTestSupport implements RewriteTest {
+class RemoveViolationCommentsTest extends AbstractPathTestSupport implements RewriteTest {
```
romani, line 31: @timtebeek , why public is not recommended ?

src/test/java/org/checkstyle/autofix/InputClassRenamer.java
```diff
@@ -51,7 +51,7 @@ private static final class ClassRenameVisitor extends JavaIsoVisitor<ExecutionCo
 
         @Override
         public J.ClassDeclaration visitClassDeclaration(J.ClassDeclaration classDecl,
-                                                        ExecutionContext executionContext) {
+                                                        ExecutionContext ctx) {
```
romani, line 54: we try to be less cryptic  and we see that being verbose is more beneficial in projects with hundreds of contributors (majority are junior level).

What is rationale to this rename ?

src/test/java/org/checkstyle/autofix/recipe/RemoveViolationCommentsTest.java
timtebeek: It's not necessary with Junit 5+, and it may lead to reuse of logic between packages, which we'd want to avoid.

src/test/java/org/checkstyle/autofix/InputClassRenamer.java
timtebeek: The rationale here is consistency with the other recipes we have; it's such a common argument that within the context of a recipe `ctx` is always the `ExecutionContext`. You can of course comment out that particular rule if you feel strongly about it.

timtebeek:
> @timtebeek is there any reliable way to get an element location (line number and column number) from the LST object? Or are there any plans to support this? That would simplify many things on our side. As for now we use some custom logic, which may be error prone - https://github.com/checkstyle/checkstyle-openrewrite-recipes/blob/main/src/main/java/org/checkstyle/autofix/PositionHelper.java

There was this earlier discontinued effort, but the code in there ought to work from what I had seen:
https://github.com/openrewrite/rewrite/pull/3829/files#diff-f66f399db6e27f51accf0f640cc8aea3384b2e26618e5a132ce120d155c5d8eaR32

Apart from that we don't have any immediate plans to support either going from coordinates to an element, or vice versa, as it's not that relevant to what we do on our end.

src/test/java/org/checkstyle/autofix/recipe/RemoveViolationCommentsTest.java
timtebeek: Reverted and disabled.

timtebeek:
Potential follow PR that can be opened & merged once this one is through:
- https://github.com/checkstyle/checkstyle-openrewrite-recipes/compare/main...timtebeek:checkstyle-openrewrite-recipes:run-recipes-on-PRs

romani (approved):
Ok to merge.
Might need squash of commits, but this project doesn't not have release notes, so it is up to @rdiachenko

rdiachenko:
@timtebeek could you resolve conflicts and squash all commits into one please? The change looks good to me, thanks a lot for contributing

romani:
as we have no special issue for this update, and this PR has a lot of comments that can explain some decision ,please name commit to start with prefix `Pull #87:  ..... `
it make sure whoever question WHY we did it in such way , easily get answer in this PR.

Issue #18, opened 2025-06-15
keep diff result in Git repository between Input and Ouput in tests


romani:
One more idea after review of PR. src/test/resources/org/checkstyle/autofix/recipe/upperell/stringandcomments/InputStringAndComments.java
It is not easy to spot a change ;).
What if we put in git repository, results of diff between Input and Output? So we will have 3 files for folder: Input, Output, diff. We don't need to automate it today. But in some future PRs, it will be beneficial. Especially beneficial for big files

rdiachenko:
@Anmol202005 .sh scripts hard to support especially when they become complex. Please reaise a draft PR and show how it looks like. If it is few lines, let's start with it. We can always iterate on a better solution

rdiachenko:
This needs to be addressed to close this issue: https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/90#issuecomment-3183583188

Anmol202005:
@rdiachenko can be closed

PR #90, opened 2025-08-13
Issue #18: Implemented git-diff generation

Anmol202005:
Fixes: #18


romani:
As initial implementation it is ok, and we should merge this.
But in long run , it will be better to run this as part of build process, to let all be generated before Pull request. In main project we use junit as launcher for file generation that should be part of commit.

Issue #80, opened 2025-08-08
Fix FinalLocalVariable Recipe for Multiple Variable Declarations


Anmol202005:
Follow up of: https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/40

The `FinalLocalVariableCheck` recipe has partial coverage and doesn't handle cases where multiple variables are declared together.

Example:
```java
int a, b, c;  // Only one variable has violation
```
Currently, cases like this where only one variable in a multi-variable declaration needs to be made final, the recipe won't process it.
Update the recipe to properly handle multiple variable declarations and apply the final modifier correctly.

Anmol202005:
@rdiachenko .

```
int a, b, c;  // Only a variable has violation
```
if in the above declaration only variable a should be declared final.
we can`t fix it.

Anmol202005:
@rdiachenko raised draft PR but requires #95 for end to end implementation.

rdiachenko:
@Anmol202005 before we take the marker approach, could you check this documentation - https://docs.openrewrite.org/authoring-recipes/multiple-visitors

I think we can try similar approach with two visitors. The first visitor will find all variable declarations that Checkstyle complains about, the second visitor will fix them. We don't need markers here, because in the first visitor we don't change the tree, so violation locations shouldn't move.

I'd like to postpone marker approach until we really need it as it is complex to do it right, and better to take simpler solution if we can.

rdiachenko:
Resolved by https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/104

PR #94, opened 2025-08-20
Issue #80: Resolved MultiVariable declaration Bug

Anmol202005:
Fixes: #80.
Requires Marker Implementation before review.

### DIFF
--- a/src/test/resources/org/checkstyle/autofix/recipe/finallocalvariable/multiplevariable/InputMultipleVariable.java
+++ b/src/test/resources/org/checkstyle/autofix/recipe/finallocalvariable/multiplevariable/InputMultipleVariable.java
@@ -16,8 +16,9 @@ public class InputMultipleVariable {
 
         String name = "John", city = "NYC";   // 2 violations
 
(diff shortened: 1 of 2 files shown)


src/test/resources/org/checkstyle/autofix/recipe/finallocalvariable/multiplevariable/InputMultipleVariable.java
```diff
@@ -0,0 +1,25 @@
...
+</module>
+*/
+
+package org.checkstyle.autofix.recipe.finallocalvariable.multiplevariable;
+
+public class InputMultipleVariable {
+
+    public void multipleDeclarations() {
+
+        int x = 10, y = 20, z = 30;           // 3 violations
```
rdiachenko, line 15: please add a test case when some of the variables are reset in the definition list, e.g.:
```
int a = 1, b = 2, c = 3; // 1 violation
...
a = 7;
b = 8;
```

So in the case above we expect 1 violation for `c` as it can be made `final`, but `a` and `b` shouldn't be final as they are set with different values later in code.

PR #104, opened 2025-08-29
Issue #80: Resolve MultiVariableDeclaration Bug

Anmol202005:
Fixes: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/80.

### DIFF
--- a/src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
+++ b/src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
@@ -19,7 +19,9 @@ package org.checkstyle.autofix.recipe;
 
 import java.nio.file.Path;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
+import java.util.UUID;
 
 import org.checkstyle.autofix.PositionHelper;
 import org.checkstyle.autofix.parser.CheckstyleViolation;
@@ -30,6 +32,8 @@ import org.openrewrite.TreeVisitor;
 import org.openrewrite.java.JavaIsoVisitor;
 import org.openrewrite.java.tree.J;
 import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.Statement;
+import org.openrewrite.marker.Marker;
 import org.openrewrite.marker.Markers;
 
 /**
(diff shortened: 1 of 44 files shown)


Anmol202005:
@rdiachenko kindly review.

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
```diff
@@ -56,17 +60,84 @@ public String getDescription() {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return new LocalVariableVisitor();
+        return new JavaIsoVisitor<>() {
```
timurt, line 63: we create this anon impl JavaIsoVisitor, after that LocalVariableVisitor and MarkViolationVisitor

It looks like all of these objects have the same generic extension, and I am curious, instead of having three visitor objects, can we have only one?

timurt: is it true, that you have to visit each variable declaration twice?

The first time to mark everything, the second time to add modifiers. If so, maybe we can have two sequential visitors

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
```diff
@@ -83,33 +154,96 @@ public J.VariableDeclarations visitVariableDeclarations(
...
+                    && varDecl.getVariables().size() > 1
+                    && !varDecl.hasModifier(J.Modifier.Type.Final)
+                    && varDecl.getTypeExpression() != null
+                    && !(getCursor().getParentTreeCursor()
+                    .getValue() instanceof J.ClassDeclaration);
+        }
+    }
+
+    public static class FinalLocalVariableMarker implements Marker {
+        private final UUID id;
```
timurt, line 233: why do we need id? couldn't find usages of this field

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
```diff
@@ -83,33 +154,96 @@ public J.VariableDeclarations visitVariableDeclarations(
...
-                    final List<J.Modifier> modifiers = new ArrayList<>();
+                if (variable.getMarkers().findFirst(FinalLocalVariableMarker.class).isPresent()) {
+                    declarations = addFinalModifier(declarations);
+                }
+            }
+            return declarations;
+        }
+
+        @Override
+        public J.Block visitBlock(J.Block block, ExecutionContext executionContext) {
```
timurt, line 165: I don't quite understand why we should specifically handle visiting blocks. My assumption was that any variable declaration node would be visited in `visitVariableDeclarations`?

Could you please provide an example of a variable declaration that won't be handled by `visitVariableDeclarations` but will be handled by `visitBlock`

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
```diff
@@ -56,17 +60,84 @@ public String getDescription() {
...
+                                                          ExecutionContext ctx) {
+                return new LocalVariableVisitor()
+                        .visitCompilationUnit(new MarkViolationVisitor()
+                                .visitCompilationUnit(cu, ctx), ctx);
+            }
+        };
     }
 
-    private final class LocalVariableVisitor extends JavaIsoVisitor<ExecutionContext> {
+    private final class MarkViolationVisitor extends JavaIsoVisitor<ExecutionContext> {
```
timurt, line 74: We definitely need more documentation around these two visitors. The logic looks cumbersome, and I think we should provide some details on how it works

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
Anmol202005: Marker interface contains abstract method like `getId()` which we need to implement.
Since `Id` is not used i have updated the marker class as :
```java
public static class FinalLocalVariableMarker implements Marker {

        public FinalLocalVariableMarker() {}

        @Override
        public UUID getId() {
            return null;
        }

        @Override
        public <M extends Marker> M withId(UUID id) {
            return null;
        }
    }
```

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
Anmol202005: This is the correct approach.
kindly see: https://docs.openrewrite.org/authoring-recipes/multiple-visitors

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
```diff
@@ -56,17 +60,89 @@ public String getDescription() {
...
+            @Override
+            public J.CompilationUnit visitCompilationUnit(J.CompilationUnit cu,
+                                                          ExecutionContext ctx) {
+                J.CompilationUnit compilationUnit = cu;
+                compilationUnit = new MarkViolationVisitor()
+                        .visitCompilationUnit(compilationUnit, ctx);
+
+                compilationUnit = new LocalVariableVisitor()
+                        .visitCompilationUnit(compilationUnit, ctx);
+
```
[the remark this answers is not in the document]
Anmol202005: I prefer logs.

rdiachenko (commented):
@Anmol202005 could you go to Checkstyle, take all test inputs from `finallocalvariable` subpackage in tests, and put them into this PR? Lots of cases are already covered by those test inputs, let's reuse them and check the recipe works expected.
image

Anmol202005:
@rdiachenko done. Kindly review

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
```diff
@@ -83,33 +164,91 @@ public J.VariableDeclarations visitVariableDeclarations(
...
+            return stmt instanceof J.VariableDeclarations varDecl
+                    && varDecl.getVariables().size() > 1
+                    && !varDecl.hasModifier(J.Modifier.Type.Final)
+                    && varDecl.getTypeExpression() != null
+                    && !(getCursor().getParentTreeCursor()
+                    .getValue() instanceof J.ClassDeclaration);
+        }
+    }
+
+    public static class FinalLocalVariableMarker implements Marker {
```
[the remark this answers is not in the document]
Anmol202005: done.

romani (approved):
ok to merge.

@Anmol202005 , do we test compilation of Input and Output files by maven compilation ?

Issue #102, opened 2025-08-26
Add support for SARIF reports


rdiachenko:
SARIF report support was introduced in Checkstyle in https://github.com/checkstyle/checkstyle/issues/8905.

SARIF is designed specifically to interchange static-analysis results across tools. That means our report-based recipe execution implementation can work for Checkstyle today and PMD/SpotBugs/... tomorrow, without rewriting parsers.

Additionally, there are available SARIF parsers, which allow to get rid of our custom implementation and support for xml parser.

rdiachenko:
fix is merged

PR #121, opened 2025-10-23
Issue #102: add support for sarif reports

rdiachenko:
Issue #102

- add support for sarif reports, so the autofix module automatically detects report type based on the violation report extension
- extend testing framework to run recipe tests against both xml and sarif reports

### DIFF
--- a/config/suppressions.xml
+++ b/config/suppressions.xml
@@ -9,8 +9,9 @@
     <suppress checks="MissingJavadocType" files=".*"/>
     <suppress checks="JavadocVariable" files=".*"/>
     <suppress checks="MissingJavadocMethod" files=".*"/>
-    <suppress checks="DesignForExtension" files="[\\/]src[\\/]test[\\/]java[\\/]org[\\/]checkstyle[\\/]autofix[\\/]recipe[\\/]AbstractRecipeTest\.java"/>
-    <suppress checks="DesignForExtension" files="[\\/]src[\\/]test[\\/]java[\\/]org[\\/]checkstyle[\\/]autofix[\\/]recipe[\\/]AbstractRecipeTestSupport\.java"/>
-    <suppress checks="DesignForExtension" files="[\\/]src[\\/]main[\\/]java[\\/]org[\\/]checkstyle[\\/]autofix[\\/]CheckstyleAutoFix\.java"/>
+
+    <!-- This project is not a library, so there is no benefit in using this check -->
+    <suppress checks="DesignForExtension" files=".*"/>
+
     <suppress checks="ClassDataAbstractionCoupling" files="[\\/]src[\\/]test[\\/]java[\\/]org[\\/]checkstyle[\\/]autofix[\\/]recipe[\\/]AbstractRecipeTestSupport\.java"/>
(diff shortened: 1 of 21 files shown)


config/suppressions.xml
```diff
@@ -9,8 +9,6 @@
     <suppress checks="MissingJavadocType" files=".*"/>
     <suppress checks="JavadocVariable" files=".*"/>
     <suppress checks="MissingJavadocMethod" files=".*"/>
-    <suppress checks="DesignForExtension" files="[\\/]src[\\/]test[\\/]java[\\/]org[\\/]checkstyle[\\/]autofix[\\/]recipe[\\/]AbstractRecipeTest\.java"/>
-    <suppress checks="DesignForExtension" files="[\\/]src[\\/]test[\\/]java[\\/]org[\\/]checkstyle[\\/]autofix[\\/]recipe[\\/]AbstractRecipeTestSupport\.java"/>
-    <suppress checks="DesignForExtension" files="[\\/]src[\\/]main[\\/]java[\\/]org[\\/]checkstyle[\\/]autofix[\\/]CheckstyleAutoFix\.java"/>
+    <suppress checks="DesignForExtension" files=".*"/>
```
rdiachenko: we are not a library, this check brings more harm then real benefits

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
```diff
@@ -56,75 +62,88 @@ protected String getPackageLocation() {
...
         final String inputPath = getInputFilePath(testCaseName);
         final String outputPath = getOutputFilePath(testCaseName);
-
         verifyOutputFile(outputPath, config);
 
-        final Path violationReportPath = runCheckstyle(inputPath, config);
-
-        final String beforeCode = readFile(getPath(inputPath));
-        final String expectedAfterCode = readFile(getPath(outputPath));
+        final Path reportPath = runCheckstyle(parser, inputPath, config);
```
romani: Using parser here sounds weird.
In code we don't not use instance at all, and we just need xml/sarif  flag .
I am not sure why we need instances of parser, but to run checkstyle we should just give enum like type  to Create xml of sarif

src/test/java/org/checkstyle/autofix/recipe/RecipeTest.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * <p>The test method must accept a single {@link org.checkstyle.autofix.parser.ReportParser}
+ * parameter, which will be automatically provided by the test framework. Each test will
+ * automatically execute twice: once with the XML report parser and once with the SARIF
+ * report parser.
+ */
+@Target(ElementType.METHOD)
+@Retention(RetentionPolicy.RUNTIME)
+@ParameterizedTest
+@MethodSource("reportParsers")
```
romani: Can we put in javadoc link to such field.
It took me some time to understand where it is

rdiachenko: done

Issue #122, opened 2025-10-23
Support multiple instances of the same Checkstyle check with different IDs


rdiachenko:
## Description

Checkstyle allows configuring the same check multiple times with different IDs and configurations. For example:

```xml
<module name="TreeWalker">
  <!-- Strict check for production code -->
  <module name="FinalLocalVariableCheck">
    <property name="id" value="strict"/>
    <property name="validateEnhancedForLoopVariable" value="true"/>
  </module>

  <!-- Lenient check for test code -->
  <module name="FinalLocalVariableCheck">
    <property name="id" value="lenient"/>
    <property name="validateEnhancedForLoopVariable" value="false"/>
  </module>
</module>
```

## Current Behavior

1. Maps checks by their `CheckstyleCheck` enum value (class name only), ignoring the `id` attribute
2. Overwrites previous configurations when multiple instances of the same check exist
3. Groups all violations from the same check type together, regardless of which check instance reported them

This means violations from `strict` and `lenient` are:
- Treated as the same check
- Fixed using only the last loaded configuration
- Cannot be distinguished or fixed differently

## Expected Behavior

1. Preserve the `id` attribute for each check instance
2. Store multiple configurations for the same check type (keyed by ID)
3. Match violations to their specific check instance using the `id` from the report
4. Apply the correct configuration when fixing violations from each instance

## Starting Point

1. Change `CheckstyleCheck` enum or create a new identity class that includes both check type and ID
2. Update `ConfigurationLoader` to preserve multiple instances with their IDs
3. Update parsers (XML/SARIF) to extract the check ID from violation reports
4. Update `CheckstyleRecipeRegistry` to match violations to the correct check instance

rdiachenko:
> The ID must be used a property, i guess i am not missing anything.

yes, I updated examples in the description

> so we need to find a way to extract the checkname from the id.

no, id is an opitonal part of check's identity, we can't extract module's name from it. We should use module's name + id (if it is present) as module's identity and map it to a corresponding configuration and recipe. This means we may need to replace existing enum with the class, which will be a compound key in the CheckstyleRecipeRegistry

PR #123, opened 2025-10-28
Issue #122: Support multiple instances of the same Checkstyle check

Anmol202005:
Fixes: #122


rdiachenko:
@Anmol202005 please create an issue in the main checkstyle repo and include examples and expected output.

Regarding:
```
    <module name="com.puppycrawl.tools.checkstyle.checks.coding.FinalLocalVariableCheck">
      <property name="id" value="com.puppycrawl.tools.checkstyle.checks.UpperEllCheck"/>
    </module>
```

I think this is more theoretical than practical case. Users are in control of their configs, and I can't imagine the case when they want to set id pointing to a different module. This can be done by some odd refactoring or mistake.

If the above case happens, what's the result of autofix execution? Will it fail or do nothing? We should just ignore such cases, because you have violation belonging to a different module id, which means if you try to execute recipe, nothing should change, right?

romani:
If people wanted to shoot in their own leg by such config, ok , it is their problem. 
There are a lot of places where you can do stupid stuff in our config, we will not support it.

src/main/java/org/checkstyle/autofix/parser/SarifReportParser.java
```diff
@@ -57,17 +55,14 @@ public List<CheckstyleViolation> parse(Path reportPath) {
...
-                    });
+                    result.add(createViolation(resultEntry.getRuleId(), resultEntry));
                 });
             }
         }
         return result;
     }
 
-    private CheckstyleViolation createViolation(CheckstyleCheck check, Result result) {
+    private CheckstyleViolation createViolation(String check, Result result) {
```
romani: is first parameter checkId ?

src/main/java/org/checkstyle/autofix/CheckstyleCheck.java
```diff
@@ -42,4 +42,10 @@ public static Optional<CheckstyleCheck> fromSource(String source) {
                 .filter(check -> check.getId().contains(source))
                 .findFirst();
     }
+
+    public static Optional<CheckstyleCheck> fromSourceExact(String source) {
```
romani: rename parameter please.

romani (changes requested):
we need some cleanup from "source" (xml attribute of checkstyle report)  in naming to checkId or checkName or ...

as we have sarif report, so better to use better term that is not xml format special

src/main/java/org/checkstyle/autofix/CheckstyleCheck.java
Anmol202005: done.

src/main/java/org/checkstyle/autofix/parser/CheckstyleViolation.java
```diff
@@ -29,14 +27,14 @@ public final class CheckstyleViolation {
 
     private final String severity;
 
-    private final CheckstyleCheck source;
+    private final String source;
```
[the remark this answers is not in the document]
Anmol202005: done.

Anmol202005:
> Please create , but please use CLI to show a problem, to let us see all details.

https://github.com/checkstyle/checkstyle/issues/18088

@romani issue created. This will make the flow straightforward here.

src/main/java/org/checkstyle/autofix/parser/CheckstyleViolation.java
```diff
@@ -29,24 +27,24 @@ public final class CheckstyleViolation {
...
-    private final CheckstyleCheck source;
+    private final String checkId;
 
     private final String message;
 
     private final Path filePath;
 
     public CheckstyleViolation(int line, int column, String severity,
-                               CheckstyleCheck source, String message, Path filePath) {
+                               String source, String message, Path filePath) {
```
romani: checkId

romani: Same for other places

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
```diff
@@ -65,18 +67,49 @@ private CheckstyleRecipeRegistry() {
...
-                    return createRecipe(entry.getValue(), config.get(entry.getKey()));
+                    final CheckConfiguration configuration =
+                            findMatchingConfiguration(entry.getKey(), config);
+                    return createRecipe(entry.getValue(), configuration);
                 })
                 .filter(Objects::nonNull)
                 .collect(Collectors.toList());
     }
 
+    private static CheckConfiguration findMatchingConfiguration(String source,
```
romani: checkId

And for all new methods.

romani (changes requested):
Few more updates required

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
```diff
@@ -65,18 +67,49 @@ private CheckstyleRecipeRegistry() {
...
+                                           Map<CheckstyleConfigModule, CheckConfiguration> config) {
+        return config.entrySet().stream()
+                .filter(configEntry -> matchesSource(configEntry.getKey(), source))
+                .map(Map.Entry::getValue)
+                .findFirst()
+                .orElse(null);
```
rdiachenko: could you return Optional and filter out absent values before trying to create recipe after calling `findMatchingConfiguration`

src/main/java/org/checkstyle/autofix/CheckstyleConfigModule.java
```diff
@@ -0,0 +1,38 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package org.checkstyle.autofix;
+
+public class CheckstyleConfigModule {
```
rdiachenko: The name is misleading, this is not a config, but just a wrapper around check, which adds id attribute from config properties. This abstraction makes the whole logic more complex, since you eventually use CheckstyleCheck.

Could you try to simplify the logic by combining both classes into one `CheckstyleCheck` and `CheckstyleConfigModule`? In reality, `id` field inside `CheckstyleCheck` is actually a fully qualified check's name

romani:
@Anmol202005 , do you have time to finish this update?
You can presume that issue in main repo is already resolved.
If any extra update required, we always follow up in new PR.

romani:
https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/127#issuecomment-3568262954 after fixing this , update for suppression is easy activity.

@Anmol202005 , fyi, fix for id support is close to merge state,  release of CS will be at last Saturday on november, so you can expect fix to be ready for you very soon.

mvnw.cmd
romani: Why it was changed,?

Anmol202005: This is a line ending normalization - the file content is identical (same checksum), but Git's metadata needed to be refreshed to match our .gitattributes rules, which caused every line to show as changed even though the actual content is unchanged. I've tried various workarounds (Git config changes, checkout options), but re-committing the file is the only reliable solution to fix the index mismatch.

romani: I just try avoid situation when next contributors with flip this file back.

Anmol202005: earlier there was no git attributes.

Anmol202005: @romani how to proceed ?

Issue #139, opened 2026-02-06
Upgrade to jdk21 and latest checkstyle version


romani:
This should unblock new checks recipe creation

Aman-Baliyan:
i am on it

romani:
Fix is merged

PR #140, opened 2026-02-06
Issue #139: Upgrade to jdk 21

Aman-Baliyan:
Issue: #139 

Upgrade jdk from version 17 to 21 and also changes the `getPackageLocation` method vissibility according to newer version of checkstyle.


romani:
Please apply pitest patch

Issue #49, opened 2025-07-17
Implement recipe for UnusedLocalVariable


rdiachenko:
Check: https://checkstyle.sourceforge.io/checks/coding/unusedlocalvariable.html#UnusedLocalVariable

Check core OpenRewrite repository and reuse existing recipes if present.

Update README with the recipe link and notes about coverage if needed.

PR #152, opened 2026-02-27
Issue #49: Implemented the recipe for UnusedLocalVariable

MonuChaudhary14:
Issue #49 

 Implemented the recipe for UnusedLocalVariable

### DIFF
--- a/src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
+++ b/src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
@@ -1,0 +1,552 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle-openrewrite-recipes: Automatically fix Checkstyle violations with OpenRewrite.
+// Copyright (C) 2025 The Checkstyle OpenRewrite Recipes Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package org.checkstyle.autofix.recipe;
+
+import java.util.ArrayList;
+import java.util.EnumSet;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.UUID;
+
+import org.checkstyle.autofix.PositionHelper;
+import org.checkstyle.autofix.parser.CheckstyleViolation;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.Comment;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.Statement;
+
+/**
+ * Fixes Checkstyle UnusedLocalVariable violations by removing local variable
+ * declarations that are never used.
+ */
+public class UnusedLocalVariable extends Recipe {
+
+    private final List<CheckstyleViolation> violations;
+
+    public UnusedLocalVariable(List<CheckstyleViolation> violations) {
+        this.violations = violations;
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "UnusedLocalVariable recipe";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Removes pure unused variables while keeping statements with side effects unchanged "
+                + "to ensure maximum code safety.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new RemoveUnusedVisitor();
+    }
+
+    private final class RemoveUnusedVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        private static final String QUOTE = "'";
+        private static final String NEW_LINE = "\n";
+        private static final String SINGLE_SPACE = " ";
+        private static final EnumSet<J.Unary.Type> UNARY_TYPES = EnumSet.of(
+                J.Unary.Type.PreDecrement,
+                J.Unary.Type.PreIncrement,
+                J.Unary.Type.PostIncrement,
+                J.Unary.Type.PostDecrement);
+
+        private final Set<UUID> namedVariablesToRemove = new HashSet<>();
+        private Set<String> removedVarNamesInMethod = new HashSet<>();
+
+        private J.CompilationUnit originalCu;
+
+        @Override
+        public J.CompilationUnit visitCompilationUnit(J.CompilationUnit compilationUnit,
+                ExecutionContext executionContext) {
+            originalCu = compilationUnit;
+            return super.visitCompilationUnit(compilationUnit, executionContext);
+        }
+
+        @Override
+        public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method,
+                ExecutionContext executionContext) {
+            final Set<String> savedRemovedNames = this.removedVarNamesInMethod;
+            this.removedVarNamesInMethod = new HashSet<>();
+            final J.MethodDeclaration result = super.visitMethodDeclaration(
+                    method, executionContext);
+            this.removedVarNamesInMethod = savedRemovedNames;
+            return result;
+        }
+
+        @Override
+        public J.VariableDeclarations visitVariableDeclarations(J.VariableDeclarations varDecl,
+                ExecutionContext executionContext) {
+            final J.VariableDeclarations visited =
+                    super.visitVariableDeclarations(varDecl, executionContext);
+
+            for (J.VariableDeclarations.NamedVariable variable : visited.getVariables()) {
+                if (isAtViolationLocation(visited, variable)) {
+                    namedVariablesToRemove.add(variable.getId());
+                    removedVarNamesInMethod.add(variable.getSimpleName());
+                }
+            }
+            return visited;
+        }
+
+        private boolean isAtViolationLocation(J.VariableDeclarations varDecl,
+                J.VariableDeclarations.NamedVariable variable) {
+            boolean matches = false;
+            final int line = PositionHelper
+                    .computeLinePosition(originalCu, varDecl, getCursor());
+
+            for (CheckstyleViolation violation : violations) {
+                final String violationPath = violation.getFilePath().toString();
+                if (violation.getLine() == line
+                        && violationPath.endsWith(originalCu.getSourcePath().toString())
+                        && violation.getMessage().contains(
+                                QUOTE + variable.getSimpleName() + QUOTE)) {
+                    matches = true;
+                    break;
+                }
+            }
+            return matches;
+        }
+
+        @Override
+        public J.Block visitBlock(J.Block block, ExecutionContext executionContext) {
+            final J.Block visited = super.visitBlock(block, executionContext);
+            return visited.withStatements(buildNewStatements(visited.getStatements()));
+        }
+
+        @Override
+        public J.Case visitCase(J.Case _case, ExecutionContext executionContext) {
+            final J.Case visited = super.visitCase(_case, executionContext);
+            return visited.withStatements(buildNewStatements(visited.getStatements()));
+        }
+
+        private List<Statement> buildNewStatements(List<Statement> statements) {
+            return rewriteStatements(statements);
+        }
+
+        private List<Statement> rewriteStatements(List<Statement> statements) {
+            final List<Statement> out = new ArrayList<>();
+            final CommentsState commentState = new CommentsState();
+            for (int index = 0; index < statements.size(); index++) {
+                final Statement stmt = statements.get(index);
+                final boolean noOutputYet = out.isEmpty();
+                final boolean replacingRemovedLead = index > 0 && noOutputYet;
+                out.addAll(rewriteOne(stmt, noOutputYet, replacingRemovedLead, commentState));
+            }
+            return out;
+        }
+
+        private List<Statement> rewriteOne(Statement stmt, boolean noOutputYet,
+                boolean replacingRemovedLead, CommentsState commentState) {
+            final RewriteResult res = produceReplacements(stmt);
+            final List<Statement> result;
+
+            if (res.getStatements().isEmpty()) {
+                commentState.absorb(stmt.getPrefix(), noOutputYet);
+                result = new ArrayList<>();
+            }
+            else if (res.isDerived()) {
+                final boolean ensureNewline = stmt instanceof J.VariableDeclarations;
+                result = applyPrefix(res.getStatements(), ensureNewline,
+                        replacingRemovedLead, commentState);
+            }
+            else {
+                result = applyPrefix(res.getStatements(), false,
+                        replacingRemovedLead, commentState);
+            }
+            return result;
+        }
+
+        private RewriteResult produceReplacements(Statement stmt) {
+            final RewriteResult result;
+            if (stmt instanceof J.VariableDeclarations varDecl) {
+                result = rewriteVariableDeclaration(varDecl);
+            }
+            else if (isOrphanedAssignment(stmt, removedVarNamesInMethod)) {
+                result = rewriteOrphanedAssignment(stmt);
+            }
+            else {
+                final List<Statement> kept = new ArrayList<>();
+                kept.add(stmt);
+                result = new RewriteResult(kept, false);
+            }
+            return result;
+        }
+
+        private RewriteResult rewriteVariableDeclaration(J.VariableDeclarations varDecl) {
+            final List<J.VariableDeclarations.NamedVariable> removed = new ArrayList<>();
+            final List<J.VariableDeclarations.NamedVariable> remaining = new ArrayList<>();
+            for (J.VariableDeclarations.NamedVariable variable : varDecl.getVariables()) {
+                if (namedVariablesToRemove.contains(variable.getId())) {
+                    removed.add(variable);
+                }
+                else {
+                    remaining.add(variable);
+                }
+            }
+
+            final RewriteResult result;
+            if (!remaining.isEmpty()) {
+                final List<Statement> stmts = new ArrayList<>();
+                stmts.add(varDecl.withVariables(remaining));
+                stmts.addAll(addInitializerStatements(removed, varDecl.getPrefix()));
+                result = new RewriteResult(stmts, false);
+            }
+            else {
+                final List<Statement> initializers = extractInitializerStatements(varDecl);
+                result = new RewriteResult(initializers, true);
+            }
+            return result;
+        }
+
+        private static RewriteResult rewriteOrphanedAssignment(Statement stmt) {
+            final List<Statement> result = new ArrayList<>();
+            final Statement sideEffectStmt = extractSideEffectFromAssignment(stmt);
+            if (sideEffectStmt != null) {
+                result.add(sideEffectStmt);
+            }
+            return new RewriteResult(result, true);
+        }
+
+        private static List<Statement> applyPrefix(List<Statement> produced,
+                boolean ensureNewline, boolean replacingRemovedLead,
+                CommentsState commentState) {
+            final List<Statement> result = new ArrayList<>(produced);
+            Statement first = result.get(0);
+
+            if (commentState.unclearedPendingWhitespace != null) {
+                final String ws = first.getPrefix().getWhitespace();
+                if (!ws.contains(NEW_LINE)) {
+                    first = first.withPrefix(first.getPrefix().withWhitespace(
+                            commentState.unclearedPendingWhitespace));
+                }
+                commentState.unclearedPendingWhitespace = null;
+            }
+
+            if (ensureNewline && first.getPrefix().getWhitespace().contains(NEW_LINE)) {
+                first = ensureNewlineInPrefix(first);
+            }
+
+            if (!commentState.getPending().isEmpty()) {
+                final String pendingWhitespace = commentState.getPendingWhitespace();
+                first = prependComments(first, commentState.getPending(), pendingWhitespace);
+                commentState.setPending(new ArrayList<>());
+            }
+
+            if (replacingRemovedLead) {
+                first = stripLeadingBlankLines(first);
+            }
+
+            first = attachTrailingComments(first, commentState.getTrailing());
+            commentState.getTrailing().clear();
+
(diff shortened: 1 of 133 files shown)


MonuChaudhary14:
@romani @Anmol202005 @rdiachenko 

Please review this PR

romani:
@MonuChaudhary14  , can you run your auto fix modules on Inputs of checkstyle for this Check to see if diff will be good and exactly that we need, no bad side effects.
Or 
is there ability to run new auto fix on some real project from web.

MonuChaudhary14:
> @MonuChaudhary14 , can you run your auto fix modules on Inputs of checkstyle for this Check to see if diff will be good and exactly that we need, no bad side effects. Or is there ability to run new auto fix on some real project from web.

Yes, I can do that. I am already testing the UseEnhancedSwitch recipe on Camunda, so I will also run this auto-fix on  Camunda and review the diff to ensure there are no unintended side effects

MonuChaudhary14:
@romani @Anmol202005 @rdiachenko 

While applying the UnusedLocalVariable recipe to the Camunda project, I found that removing unused variables with side-effecting initializers (like constructors or method calls) can break application logic. To prevent this, I updated the recipe to only remove variables with safe initializers (literals, etc.)

However, I've encountered a technical limitation in our integration test framework:
To verify that an unsafe variable is correctly preserved, it must remain in the output file.
But our test requires all output files to be violation free
Since the preserved variable is still unused, Checkstyle flags it as a violation in the output, causing the test to fail.

If I try to suppress the violation in the input to make the output pass, the recipe skips the variable entirely, so isSafeToRemove never actually gets to evaluate that specific case. I've tried a few different workarounds in the recipe code to trigger this path, but nothing worked

Because of this, I have a few surviving Pitest mutations in the safety check that are effectively impossible to kill with our integration tests alone. Given this limitation, would it be okay to suppress these specific mutations in the PIT configuration?

https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/152/changes#diff-b15af6351f624c8a4e69d409831b64b53338cabae3c8dabccc21a52b2f404cb2R14

MonuChaudhary14:
> @romani @Anmol202005 @rdiachenko
> 
> While applying the UnusedLocalVariable recipe to the Camunda project, I found that removing unused variables with side-effecting initializers (like constructors or method calls) can break application logic. To prevent this, I updated the recipe to only remove variables with safe initializers (literals, etc.)
> 
> However, I've encountered a technical limitation in our integration test framework: To verify that an unsafe variable is correctly preserved, it must remain in the output file. But our test requires all output files to be violation free Since the preserved variable is still unused, Checkstyle flags it as a violation in the output, causing the test to fail.
> 
> If I try to suppress the violation in the input to make the output pass, the recipe skips the variable entirely, so isSafeToRemove never actually gets to evaluate that specific case. I've tried a few different workarounds in the recipe code to trigger this path, but nothing worked
> 
> Because of this, I have a few surviving Pitest mutations in the safety check that are effectively impossible to kill with our integration tests alone. Given this limitation, would it be okay to suppress these specific mutations in the PIT configuration?
> 
> https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/152/changes#diff-b15af6351f624c8a4e69d409831b64b53338cabae3c8dabccc21a52b2f404cb2R14

@romani ?

romani:
>But our test requires all output files to be violation free
Since the preserved variable is still unused, Checkstyle flags it as a violation in the output, causing the test to fail.

this means there is defect in Checkstyle Check, it should not point to this variable as not used, please create issue on this in main reporsitory.
meanwhile lets do suppress with reference in comment "until xxxxxx".

MonuChaudhary14:
@romani 

Please check the issue

https://github.com/checkstyle/checkstyle/issues/19260

romani:
https://github.com/checkstyle/checkstyle/issues/19260#issuecomment-4117711594 

I think we need to improve auto-fix, it should not remove whole statements but remove initialization of variable from statement, as method can give side effect.
detection of side effect is over complicated.

MonuChaudhary14:
> [checkstyle/checkstyle#19260 (comment)](https://github.com/checkstyle/checkstyle/issues/19260#issuecomment-4117711594)
> 
> I think we need to improve auto-fix, it should not remove whole statements but remove initialization of variable from statement, as method can give side effect. detection of side effect is over complicated.

done

MonuChaudhary14:
@romani @Anmol202005 
I have made all the changes and tested it on dubbo and jenkins,

Here is the report for it
https://gist.githubusercontent.com/MonuChaudhary14/d248c07e5f9c6d18718f70f975d1ff68/raw/1739423b67baae6e4d04cafd331e7864547202ac/Unusedlocalvariable.diff

MonuChaudhary14:
@romani 
Ping

MonuChaudhary14:
> @romani Ping

ping

MonuChaudhary14:
@romani @Anmol202005 @rdiachenko 

Ping

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
```diff
@@ -0,0 +1,342 @@
...
+            return visited.withStatements(buildNewStatements(visited.getStatements()));
+        }
+
+        @Override
+        public J.Case visitCase(J.Case _case, ExecutionContext executionContext) {
+            final J.Case visited = super.visitCase(_case, executionContext);
+            return visited.withStatements(buildNewStatements(visited.getStatements()));
+        }
+
+        private boolean isAtViolationLocation(J.VariableDeclarations varDecl,
```
sabaka, line 116: Could you move method to be right after visitVariableDeclarations please? it will be easier to read this way.

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
```diff
@@ -0,0 +1,342 @@
...
+            }
+            return result;
+        }
+
+        private void handleVariableDeclaration(J.VariableDeclarations varDecl,
+                List<Statement> newStatements, List<Comment> pendingComments,
+                List<Comment> pendingTrailingComments) {
+            final List<J.VariableDeclarations.NamedVariable> removed = new ArrayList<>();
+            final List<J.VariableDeclarations.NamedVariable> remaining =
+                    org.openrewrite.internal.ListUtils.map(varDecl.getVariables(),
```
sabaka: Could be an import

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
```diff
@@ -0,0 +1,342 @@
...
+                final Expression rhs = unwrap(assignment.getAssignment());
+                if (rhs instanceof Statement rhsStmt) {
+                    result = rhsStmt.withPrefix(stmt.getPrefix());
+                }
+            }
+            return result;
+        }
+
+        private void handleVariableDeclaration(J.VariableDeclarations varDecl,
+                List<Statement> newStatements, List<Comment> pendingComments,
```
sabaka: Is it possible to change this method to return `List<Statement>` so in-place modification (this method does) isn't mixed with simple addition to the list in called method?

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
```diff
@@ -0,0 +1,342 @@
...
+                    newStatements.add(initStatements.get(idx));
+                }
+            }
+            else {
+                transferComments(varDecl, newStatements,
+                        pendingComments, pendingTrailingComments);
+            }
+        }
+
+        private static void addInitializerStatements(
```
sabaka: Please move it to be after `handleVariableDeclaration` to be more readable.

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
```diff
@@ -0,0 +1,342 @@
...
+                result = unwrap((Expression) parens.getTree());
+            }
+            return result;
+        }
+
+        private static boolean isStatementExpression(Expression expr) {
+            final boolean isUnaryIncremental;
+            if (expr instanceof J.Unary unary) {
+                final J.Unary.Type type = unary.getOperator();
+                isUnaryIncremental = type == J.Unary.Type.PreIncrement
```
sabaka: Can be something like:
```java
private static final EnumSet<J.Unary.Type> UNARY_TYPES = EnumSet.of(J.Unary.Type.PreDecrement, J.Unary.Type.PreIncrement, J.Unary.Type.PostIncrement, J.Unary.Type.PostDecrement);
...
isUnaryIncremental = UNARY_TYPES.contains(type);
```
to make condition simpler

sabaka (commented):
Left a few comments but haven't finished yet

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
MonuChaudhary14: done

MonuChaudhary14:
@sabaka 
I have made some changes in the recipe, because there was a problem in the previous one, 
can you please check the new appraoch

MonuChaudhary14:
@sabaka 
Is there any changes needed

MonuChaudhary14:

Please review this PR, and tell me if anything more needed to do in this PR

romani:
@MonuChaudhary14 , how can we run this on some real project to see how it works on wild code ?
Merged PR for some project is best form to prove that implementation is GOOD.

src/test/resources/org/checkstyle/autofix/recipe/unusedlocalvariable/killremainingmutations/DiffKillRemainingMutations.diff
```diff
@@ -0,0 +1,118 @@
...
+-
+-        int unused = 2; // violation 'Unused named local variable'
++        int used = 1; /* block comment */
+         System.out.println(used);
+     }
+ 
+     public void twoComments() {
+-        int used = 1;
+-        // comment 1
++        int used = 1; // comment 1
```
romani: let avoid placing comments as trailing comments, usage of trailing comments is not good style in many cases, not always.
lets simply remove unused variable

src/test/resources/org/checkstyle/autofix/recipe/unusedlocalvariable/killremainingmutations/DiffKillRemainingMutations.diff
```diff
@@ -0,0 +1,118 @@
...
+@@ -25,10 +24,7 @@
+         System.out.println(arr[0]);
+     }
+ 
+-    public void mergeIntoExistingComments() {
+-        // comment on unused
+-        int unused = 1; // violation 'Unused named local variable'
+-        int unused2 = 2; // violation 'Unused named local variable'
++    public void mergeIntoExistingComments() { // comment on unused
+         // existing comment
```
romani, line 25: lets keep comments on lines they were.

src/test/resources/org/checkstyle/autofix/recipe/unusedlocalvariable/killremainingmutations/DiffKillRemainingMutations.diff
```diff
@@ -0,0 +1,118 @@
...
++            /* comment */ System.out.println("next");
+         }
+     }
+ 
+     public void multipleKeptStatements() {
+-        int unused = 0; // violation 'Unused named local variable'
+-        /* c1 */ unused = 1;
+-        System.out.println("A");
++        /* c1 */ System.out.println("A");
+         System.out.println("B");
```
romani:
```
/* c1 */
System.out.println("A");
```

romani (changes requested):
items:

MonuChaudhary14:
> @MonuChaudhary14 , how can we run this on some real project to see how it works on wild code ? Merged PR for some project is best form to prove that implementation is GOOD.

Okay, I will do that. I’ll try running this on multiple repos as well, if possible.

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
```diff
@@ -0,0 +1,430 @@
...
+        @Override
+        public J.Case visitCase(J.Case _case, ExecutionContext executionContext) {
+            final J.Case visited = super.visitCase(_case, executionContext);
+            return visited.withStatements(buildNewStatements(visited.getStatements()));
+        }
+
+        private List<Statement> buildNewStatements(List<Statement> statements) {
+            final List<Statement> newStatements = new ArrayList<>();
+            final CommentsState state = new CommentsState();
+            boolean isNewStatementsEmpty = true;
```
sabaka: I don't think you need this variable. You can make newStmts.isEmpty() call when you need this value.

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
```diff
@@ -0,0 +1,430 @@
...
+            for (int index = 0; index < statements.size(); index++) {
+                final Statement stmt = statements.get(index);
+                final boolean isFirstOriginal = index == 0;
+
+                if (stmt instanceof J.VariableDeclarations varDecl) {
+                    final List<Statement> newStmts = handleVariableDeclaration(varDecl,
+                            isNewStatementsEmpty, state, isFirstOriginal);
```
sabaka: You pass `isNewStatementsEmpty` deep in method call chain and use it to remove blank lines. Is it possible to revert stripping logic a bit and do it here after processing? Then you wouldn't need to pass the var and I think code will be a bit easier to read

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
```diff
@@ -0,0 +1,430 @@
...
+
+        @Override
+        public J.Case visitCase(J.Case _case, ExecutionContext executionContext) {
+            final J.Case visited = super.visitCase(_case, executionContext);
+            return visited.withStatements(buildNewStatements(visited.getStatements()));
+        }
+
+        private List<Statement> buildNewStatements(List<Statement> statements) {
+            final List<Statement> newStatements = new ArrayList<>();
+            final CommentsState state = new CommentsState();
```
sabaka: could you name it commentState please?

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
```diff
@@ -0,0 +1,430 @@
...
+                    result = removedNames.contains(id.getSimpleName());
+                }
+            }
+            return result;
+        }
+
+        private static Statement extractSideEffectFromAssignment(Statement stmt) {
+            Statement result = null;
+            if (stmt instanceof J.Assignment assignment) {
+                final Expression rhs = unwrap(assignment.getAssignment());
```
sabaka: could name be more descriptive than `rhs`?

romani:
@MonuChaudhary14 , conflict.

src/test/resources/org/checkstyle/autofix/recipe/unusedlocalvariable/killtransfercommentmutations/DiffKillTransferCommentMutations.diff
```diff
@@ -0,0 +1,47 @@
...
+         // comment before orphaned
+-        a = 10;
+         System.out.println(b);
+     }
+ 
+     public void trailingCommentBeforeOrphaned() {
+-        int a = 1; // violation 'Unused named local variable'
+-        int b = 2; /* same line comment */ a = 10;
+-        System.out.println(b);
++        int b = 2; /* same line comment */ System.out.println(b);
```
romani: Can we keep prinntln on its own line?

romani (changes requested):
Items

src/test/resources/org/checkstyle/autofix/recipe/unusedlocalvariable/killtransfercommentmutations/DiffKillTransferCommentMutations.diff
MonuChaudhary14: I think I missed this. I’ll look into it and get back to you soon.

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
```diff
@@ -0,0 +1,493 @@
...
+            return visited.withStatements(buildNewStatements(visited.getStatements()));
+        }
+
+        @Override
+        public J.Case visitCase(J.Case _case, ExecutionContext executionContext) {
+            final J.Case visited = super.visitCase(_case, executionContext);
+            return visited.withStatements(buildNewStatements(visited.getStatements()));
+        }
+
+        private List<Statement> buildNewStatements(List<Statement> statements) {
```
sabaka, line 147: Sorry, I know it was a bit back and forth here, but I still think this method is too complex and challenging to follow. Let's try to break it down so responsibility is cleaner.

Here is some not-too-detailed pseudocode with the idea.

```

  rewriteStatements(statements):
      out          = []
      commentState = CommentsState()   # same as today: pending + trailing
                                       # comments salvaged from removed stmts
      for i, stmt in statements:
          noOutputYet          = out.isEmpty()
          replacingRemovedLead = (i > 0) and noOutputYet
          out.addAll(rewriteOne(stmt, noOutputYet,
                                replacingRemovedLead, commentState))
      return out

  rewriteOne(stmt, noOutputYet, replacingRemovedLead, commentState):
      # 1. peel prefix off
      prefix = stmt.prefix

      # 2. type-pure rewrite: no comment handling inside
      produced = produceReplacements(stmt)

      # 3. fully removed? carry comments forward, emit nothing
      if produced.isEmpty():
          commentState.absorb(prefix, noOutputYet)
          return []

      # 4. reattach carried-over comments + whitespace
      return applyPrefix(produced, replacingRemovedLead, commentState)

  produceReplacements(stmt):
      # dispatch on statement type — each branch only knows about its own shape
      if stmt is VariableDeclarations:
          return rewriteVariableDeclaration(stmt)   # slimmed var decl +
                                                    # extracted side-effect inits
      if stmt is orphaned assignment / unary:
          return rewriteOrphanedAssignment(stmt)    # the salvaged RHS, or []
      return [stmt]                                 # kept as-is


```

Let me know if this makes sense or if I'm missing something.

MonuChaudhary14: I’ll check this approach. Most of the complexity in the method came from trying to kill the surviving mutations, and this section was the hardest part to make mutation-free.

That said, your breakdown does make the responsibilities much clearer compared to the current flow.I’ll see if I can restructure it in this direction without bringing those mutations back.

src/test/resources/org/checkstyle/autofix/recipe/unusedlocalvariable/multiplesideeffects/DiffMultipleSideEffects.diff
```diff
@@ -0,0 +1,19 @@
...
++public class OutputMultipleSideEffects {
+ 
+     public void method() {
+         int a = 0;
+         // Comment before multiple side effects
+-        int unused1 = ++a, unused2 = a++; // 2 violations
++        ++a;
++        // Comment before multiple side effects
++        a++;
+         System.out.println(a);
```
romani, line 15: should be:
```
int a = 0;
// Comment before multiple side effects
++a; a++;
```

romani (changes requested):
items:

src/test/resources/org/checkstyle/autofix/recipe/unusedlocalvariable/killmutations/DiffKillMutations.diff
```diff
@@ -0,0 +1,451 @@
...
+     }
+ 
+     public void partialWithTrailingCommentAndExtracted() {
+-        int keep = 1; int unused1 = 2; /* trailing */ // violation 'Unused named local variable'
+-        int unused2 = getString().length(); // violation 'Unused named local variable'
++        int keep = 1;
++        /* trailing */
++        getString().length();
+         System.out.println(keep);
+     }
```
MonuChaudhary14: @romani @sabaka 

I want to ask you opinion on this,
When unused1 is deleted, the recipe attaches its trailing comment /* trailing */ to the next statement (getString()) so it doesn't get wrongly linked to keep = 1. Since getString() gets moved to a new line, the comment naturally goes right above it on its own line. This avoids confusion, ensuring it's clear the comment is not related to keep = 1. Should I keep it like this or do we want to change this?

romani: `int keep = 1; int unused1 = 2; /* trailing */`
Should become 
`int keep = 1; /* trailing */`

romani: We don't know if comment belong to Removed variable. User will do extra work to clean up code

romani (approved):
I am tired on this PR. Good for me

src/main/java/org/checkstyle/autofix/recipe/UnusedLocalVariable.java
sabaka: Cool, looks much better. Please resolve conflicts and let's merge it

MonuChaudhary14: > Cool, looks much better. Please resolve conflicts and let's merge it

This will take some time, when i rebase this, then i found out that there is one illegal import of 
import org.openrewrite.internal.ListUtils;
let me fix this and get to you soon

romani:
Conflict

MonuChaudhary14:
> Conflict

@romani 

I have to make some changes here. After rebasing, I noticed that we are using an internal OpenRewrite API:
https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/152/changes#diff-c4f1a092ef22326b021c716fe32f8e8ccd6d8297bdaea9b47a7c7f8083e69718R32

When I update this, it introduces around 30 mutations.

If you don’t want to review the whole PR again, we can suppress these for now and handle the changes in a follow-up PR.

MonuChaudhary14:
@romani @sabaka 

Currently, i have removed the use of internal api, but for now i have suppressed the file length for two input files, we can update this on this issue
https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/202

romani:
conflict

Issue #95, opened 2025-08-20
Replace position-based recipe implementation with marker-based approach


Anmol202005:
Currently, our recipe implementation is position-based - recipes are applied when violations are detected at specific positions. However, this creates issues because violation positions change after each recipe is implemented.

Replace the position-based approach with a marker-based system:

- Before implementing fixes: Visit each violation and attach a unique marker to it
- During recipe implementation: Work with markers instead of positions, making the system more robust to position changes

romani:
@Aman-Baliyan, can you help us investigate this issue?

Aman-Baliyan:
ok sir, I’m currently investigating the issue and going through the relevant parts of the codebase to understand the logic properly. I’ll need a bit of time to analyze it, but I’ll update you with my findings soon.

Aman-Baliyan:
@romani sir, i have investigated the issue and confirm that there is a big problem with the position based approach because after removal of a node from lst changes the line and column position of another node.
So  the marker approach is needed and i have a design idea for this approach - 

Step 1 – Marker Recipes

We introduce dedicated marker recipes that attach a specific ViolationMarker to the LST before any transformations occur.
Each fixing recipe  would have its own marker recipe.
This is necessary because different violations may require different logic to locate the correct node (especially when the violation refers to punctuation, whitespace, or other non-node elements).
The responsibility of these recipes is strictly identification and marker attachment — no modifications.

Step 2 – Fixing Recipes

Each fixing recipe would:
Traverse the LST looking only for its corresponding ViolationMarker.
Apply the transformation directly to the marked node.
This cleanly separates detection from transformation.

And the main change is that all marker recipe will runs before any fixing recipe. By creating separate marker recipe it is easier to track violations.
Is i am thinking in correct direction?

rdiachenko:
@Aman-Baliyan good first step. We should aim to have one "super" recipe or corresponging to markers logic to run before all checkstyle recipes. We don't want to end up writing a separate marker recipe every time we create a new fixing recipe. We have infromation about checkstyle violations in the report. Could you expore deeper with some mock code examples of how this may look like. Also check the openrewrite docs for suggeted approach for markers.

MonuChaudhary14:
@romani @rdiachenko @Anmol202005 

I am also working on this issue.The solution I think we should use ScanningRecipe which is already present in the openrewrite. 

The ScanningRecipe have two phases:
[...]

PR #153, opened 2026-02-28
Issue #95: Replace position-based recipe implementation with marker-based approach

MonuChaudhary14:
Issue #95 

Replace position-based recipe implementation with marker-based approach

### DIFF
--- a/src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
+++ b/src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
@@ -17,14 +17,17 @@
 
 package org.checkstyle.autofix;
 
+import java.util.ArrayList;
 import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
+import java.util.Optional;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+import org.checkstyle.autofix.marker.ViolationMarkerRecipe;
 import org.checkstyle.autofix.parser.CheckConfiguration;
 import org.checkstyle.autofix.parser.CheckstyleViolation;
 import org.checkstyle.autofix.recipe.AnnotationOnSameLine;
@@ -55,15 +58,16 @@ public final class CheckstyleRecipeRegistry {
             CheckConfiguration, Recipe>> RECIPE_MAP_WITH_CONFIG =
             new EnumMap<>(CheckFullName.class);
 
+    private static final EnumMap<CheckFullName, Supplier<Recipe>>
+            RECIPE_MAP_NO_VIOLATIONS = new EnumMap<>(CheckFullName.class);
+
     static {
         RECIPE_MAP.put(CheckFullName.AVOID_STAR_IMPORT, AvoidStarImport::new);
         RECIPE_MAP.put(CheckFullName.EMPTY_STATEMENT, EmptyStatement::new);
         RECIPE_MAP.put(CheckFullName.CONSTRUCTORS_DECLARATION_GROUPING,
             ConstructorsDeclarationGrouping::new);
-        RECIPE_MAP.put(CheckFullName.FINAL_CLASS, FinalClass::new);
         RECIPE_MAP.put(CheckFullName.UPPER_ELL, UpperEll::new);
         RECIPE_MAP.put(CheckFullName.HEX_LITERAL_CASE, HexLiteralCase::new);
-        RECIPE_MAP.put(CheckFullName.FINAL_LOCAL_VARIABLE, FinalLocalVariable::new);
         RECIPE_MAP_WITH_CONFIG.put(CheckFullName.HEADER, Header::new);
         RECIPE_MAP_WITH_CONFIG.put(CheckFullName.NEWLINE_AT_END_OF_FILE, NewlineAtEndOfFile::new);
         RECIPE_MAP.put(CheckFullName.NUMERICAL_PREFIXES_INF_SUF_CASE,
@@ -75,6 +79,8 @@ public final class CheckstyleRecipeRegistry {
         RECIPE_MAP.put(CheckFullName.MISSING_OVERRIDE, MissingOverride::new);
         RECIPE_MAP.put(CheckFullName.MISSING_DEPRECATED, MissingDeprecated::new);
         RECIPE_MAP.put(CheckFullName.UNUSED_IMPORT, UnusedImports::new);
+        RECIPE_MAP_NO_VIOLATIONS.put(CheckFullName.FINAL_CLASS, FinalClass::new);
+        RECIPE_MAP_NO_VIOLATIONS.put(CheckFullName.FINAL_LOCAL_VARIABLE, FinalLocalVariable::new);
     }
 
     private CheckstyleRecipeRegistry() {
@@ -92,35 +98,41 @@ public final class CheckstyleRecipeRegistry {
      */
     public static List<Recipe> getRecipes(List<CheckstyleViolation> violations,
                                           Map<CheckstyleCheck, CheckConfiguration> config) {
-        return violations.stream()
-                .collect(Collectors.groupingBy(CheckstyleViolation::getSource))
-                .entrySet()
-                .stream()
-                .map(entry -> {
-                    return createRecipe(entry.getValue(), config.get(entry.getKey()));
-                })
-                .filter(Objects::nonNull)
-                .collect(Collectors.toList());
+        final List<Recipe> recipes = new ArrayList<>();
+
+        recipes.add(new ViolationMarkerRecipe(violations));
+
+        final Map<CheckstyleCheck, List<CheckstyleViolation>> groupedViolations =
+                violations.stream()
+                        .collect(Collectors.groupingBy(CheckstyleViolation::getSource));
+
+        for (Map.Entry<CheckstyleCheck, List<CheckstyleViolation>> entry
+                : groupedViolations.entrySet()) {
+            final CheckstyleCheck check = entry.getKey();
+
+            Optional.ofNullable(config.get(check))
+                    .ifPresent(checkConfig -> addRecipe(recipes, checkConfig, entry.getValue()));
+        }
+
+        return recipes;
     }
 
-    private static Recipe createRecipe(List<CheckstyleViolation> violations,
-                                       CheckConfiguration checkConfig) {
-        Recipe result = null;
-        if (checkConfig != null) {
+    private static void addRecipe(List<Recipe> recipes, CheckConfiguration checkConfig,
+                                  List<CheckstyleViolation> violations) {
+        final CheckFullName checkName = checkConfig.getCheckName();
 
-            final CheckFullName checkName = checkConfig.getCheckName();
+        Optional.ofNullable(RECIPE_MAP_NO_VIOLATIONS.get(checkName))
+                .ifPresent(factory -> recipes.add(factory.get()));
 
-            final BiFunction<List<CheckstyleViolation>, CheckConfiguration,
-                    Recipe> configRecipeFactory = RECIPE_MAP_WITH_CONFIG.get(checkName);
+        Optional.ofNullable(RECIPE_MAP_WITH_CONFIG.get(checkName))
+                .ifPresent(factory -> {
+                    recipes.add(factory.apply(violations, checkConfig));
+                });
 
-            if (configRecipeFactory != null) {
-                result = configRecipeFactory.apply(violations, checkConfig);
-            }
-            else {
-                result = RECIPE_MAP.get(checkName).apply(violations);
-            }
-        }
-        return result;
+        Optional.ofNullable(RECIPE_MAP.get(checkName))
+                .ifPresent(factory -> {
+                    recipes.add(factory.apply(violations));
+                });
     }
 
 }
--- a/src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
+++ b/src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
@@ -1,0 +1,345 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle-openrewrite-recipes: Automatically fix Checkstyle violations with OpenRewrite.
+// Copyright (C) 2025 The Checkstyle OpenRewrite Recipes Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package org.checkstyle.autofix.marker;
+
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.UUID;
+
+import org.checkstyle.autofix.CheckFullName;
+import org.checkstyle.autofix.parser.CheckstyleViolation;
+import org.openrewrite.Cursor;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.PrintOutputCapture;
+import org.openrewrite.ScanningRecipe;
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.marker.Marker;
+
+public class ViolationMarkerRecipe extends ScanningRecipe<Accumulator> {
+
+    private static final String DISPLAY_NAME = "Checkstyle violation marker";
+    private static final String DESCRIPTION =
+            "Marks AST nodes that correspond to Checkstyle violations.";
+
+    private static final MarkersApplied APPLIED_MARKER = new MarkersApplied(Tree.randomId());
+
+    private final List<CheckstyleViolation> violations;
+
+    public ViolationMarkerRecipe(List<CheckstyleViolation> violations) {
+        this.violations = new ArrayList<>(violations);
+    }
+
+    @Override
+    public String getDisplayName() {
+        return DISPLAY_NAME;
+    }
+
+    @Override
+    public String getDescription() {
+        return DESCRIPTION;
+    }
+
+    @Override
+    public Accumulator getInitialValue(ExecutionContext executionContext) {
+        return new Accumulator();
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getScanner(Accumulator acc) {
+        return new ScannerVisitor(acc);
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor(Accumulator acc) {
+        return new MarkerVisitor(acc);
+    }
+
+    private final class ScannerVisitor extends JavaIsoVisitor<ExecutionContext> {
+        private final Accumulator acc;
+
+        ScannerVisitor(Accumulator acc) {
+            this.acc = acc;
+        }
+
+        @Override
+        public J.CompilationUnit visitCompilationUnit(J.CompilationUnit compUnit,
+                                                      ExecutionContext executionContext) {
+            final J.CompilationUnit result;
+            if (compUnit.getMarkers().findFirst(MarkersApplied.class).isPresent()) {
+                result = compUnit;
+            }
+            else {
+                final Path sourcePath = compUnit.getSourcePath();
+                final List<CheckstyleViolation> fileViolations = violations.stream()
+                        .filter(violation -> Objects.equals(violation.getFilePath(), sourcePath))
+                        .toList();
+
+                if (!fileViolations.isEmpty()) {
+                    processFileViolations(compUnit, sourcePath, fileViolations);
+                }
+                result = compUnit;
+            }
+            return result;
+        }
+
+        private void processFileViolations(J.CompilationUnit compilationUnit, Path sourcePath,
+                                           List<CheckstyleViolation> fileViolations) {
+            final Map<UUID, Range> nodeRanges = new LinkedHashMap<>();
+            final Map<UUID, UUID> parentMap = new HashMap<>();
+            final Map<UUID, Tree> nodeTrees = new HashMap<>();
+            final TreeVisitor<?, PrintOutputCapture<TreeVisitor<?, ?>>> printer =
+                    compilationUnit.printer(new Cursor(null, "root"));
+
+            final BoundingBoxCapture capture = new BoundingBoxCapture(
+                    printer, nodeRanges, parentMap, nodeTrees);
+            printer.visit(compilationUnit, capture, getCursor().getParentOrThrow());
+
+            final Map<UUID, List<CheckstyleViolationMarker>> markersToAdd = new HashMap<>();
+            for (CheckstyleViolation violation : fileViolations) {
+                final Map<UUID, List<CheckstyleViolationMarker>> result = findSmallestNode(
+                        violation, nodeRanges, parentMap, nodeTrees);
+                for (Map.Entry<UUID, List<CheckstyleViolationMarker>> entry : result.entrySet()) {
+                    markersToAdd.computeIfAbsent(entry.getKey(), id -> new ArrayList<>())
+                            .addAll(entry.getValue());
+                }
+            }
+
+            if (!markersToAdd.isEmpty()) {
+                acc.putByFile(sourcePath, markersToAdd);
+            }
+        }
+
+        private Map<UUID, List<CheckstyleViolationMarker>> findSmallestNode(
+                CheckstyleViolation violation, Map<UUID, Range> nodeRanges,
+                Map<UUID, UUID> parentMap, Map<UUID, Tree> nodeTrees) {
+            UUID smallestNodeId = null;
+            int minLines = Integer.MAX_VALUE;
+            int minCols = Integer.MAX_VALUE;
+
+            for (Map.Entry<UUID, Range> entry : nodeRanges.entrySet()) {
+                final Range range = entry.getValue();
+                if (encloses(range, violation)) {
+                    final int lines = range.endLine() - range.startLine();
+                    final int cols = range.endCol() - range.startCol();
+                    if (lines < minLines || lines == minLines && cols <= minCols) {
+                        minLines = lines;
+                        minCols = cols;
+                        smallestNodeId = entry.getKey();
+                    }
+                }
+            }
+
+            final Map<UUID, List<CheckstyleViolationMarker>> result = new HashMap<>();
+            if (smallestNodeId != null) {
+                final UUID targetId = resolveTargetNode(smallestNodeId, violation,
+                        parentMap, nodeTrees);
+                result.computeIfAbsent(targetId, nodeId -> new ArrayList<>())
+                        .add(new CheckstyleViolationMarker(Tree.randomId(), violation));
+            }
+            return result;
+        }
+
+        private UUID resolveTargetNode(UUID startNodeId, CheckstyleViolation violation,
+                                       Map<UUID, UUID> parentMap, Map<UUID, Tree> nodeTrees) {
+            UUID resultId = startNodeId;
+            final CheckFullName checkFullName = violation.getSource().checkName();
+            if (checkFullName != null) {
+                final Class<? extends Tree> targetType = switch (checkFullName) {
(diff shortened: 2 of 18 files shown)


Anmol202005:
@MonuChaudhary14 Please update only one recipe in this PR. This should help pass the PITest checks. We can update the remaining recipes in separate PRs.

MonuChaudhary14:
@romani @Anmol202005 @rdiachenko 

I have made the changes, replaced the position based with marker based approach

Anmol202005:
Thanks for the update, @MonuChaudhary14  appreciate the effort here. However, I have some architectural concerns with the current approach.

The core issue: Markers should be registered upfront  before any fix recipe runs  rather than added per-recipe at execution time.
In the current implementation, each recipe adds its own markers as it executes. Since recipes run sequentially within CheckstyleAutoFix, any fix applied by an earlier recipe shifts the source positions of subsequent violations. This means later recipes are still searching against stale/incorrect positions, which is the exact problem we set out to solve.
The intended design should work like this:

- CheckstyleAutoFix iterates over all violations and adds markers for all of them upfront, in a single pre-processing pass  before any individual fix recipe is invoked.
- Each child recipe then searches by marker rather than by source position, making it position-independent and resilient to edits made by preceding recipes.

This separation of concerns (mark-all-first → fix-by-marker) is what gives us correctness guarantees across the full fix pipeline.

Suggested next steps:

- Move the marker registration logic into CheckstyleAutoFix as a dedicated pre-pass over the full violation list.
- Update each individual recipe to locate the relevant AST node via its marker rather than by line/column position.

MonuChaudhary14:
> Thanks for the update, @MonuChaudhary14 appreciate the effort here. However, I have some architectural concerns with the current approach.
> 
> The core issue: Markers should be registered upfront before any fix recipe runs rather than added per-recipe at execution time. In the current implementation, each recipe adds its own markers as it executes. Since recipes run sequentially within CheckstyleAutoFix, any fix applied by an earlier recipe shifts the source positions of subsequent violations. This means later recipes are still searching against stale/incorrect positions, which is the exact problem we set out to solve. The intended design should work like this:
> 
> * CheckstyleAutoFix iterates over all violations and adds markers for all of them upfront, in a single pre-processing pass  before any individual fix recipe is invoked.
> * Each child recipe then searches by marker rather than by source position, making it position-independent and resilient to edits made by preceding recipes.
> 
> This separation of concerns (mark-all-first → fix-by-marker) is what gives us correctness guarantees across the full fix pipeline.
> 
> Suggested next steps:
> 
> * Move the marker registration logic into CheckstyleAutoFix as a dedicated pre-pass over the full violation list.
> * Update each individual recipe to locate the relevant AST node via its marker rather than by line/column position.

done

romani:
@MonuChaudhary14 , please rebase.


@Anmol202005, please review to share your opinion on this proposal.

Anmol202005:
@MonuChaudhary14 please add a test with two violations in the same line. 
ideally the test should fail without these changes and pass with the marker implementation.

MonuChaudhary14:
> @MonuChaudhary14 please add a test with two violations in the same line. ideally the test should fail without these changes and pass with the marker implementation.

I have added a new test for finalClass , but it is passing with position based approach because it is only considering line number, so now i am trying to run two different types of recipe on one input, will update you soon

MonuChaudhary14:
@Anmol202005 
I have added the test case, which is breaking on position based approach
https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/153/changes#diff-d26408929822afdea55179c04c325500c7907f00566f79280f71d9af18b90dc8R1

while finding the test case, i have also changed the upper_ell and finalLocalvariable to marker based approach, if you want i can revert them back

Anmol202005:
> while finding the test case, i have also changed the upper_ell and finalLocalvariable to marker based approach, if you want i can revert them back

kindly revert changes for `upper_ell`  and `finalClass`,
also kindly make the CI happy

MonuChaudhary14:
> > while finding the test case, i have also changed the upper_ell and finalLocalvariable to marker based approach, if you want i can revert them back
> 
> kindly revert changes for `upper_ell` and `finalClass`, also kindly make the CI happy

I have made the CI green, also i have revert the changes of upper_ell, 
the test file is using both finalClass and finalLocalVariable violations, so i have not reverted the changes of finalClass

MonuChaudhary14:
@rdiachenko 

Please review this

MonuChaudhary14:
@rdiachenko 
Please review this PR

MonuChaudhary14:
Please review this PR, and tell me if any changes needed

romani:
@MonuChaudhary14 , please resove conflicts.

romani:
@MonuChaudhary14, please invest time to cover this unrelated mutation problem, Ithis blocks and annoy to much, we need workaround

romani:
Rebased in web, to see pitest results

romani:
@MonuChaudhary14 , there is confict

MonuChaudhary14:
> @MonuChaudhary14 , there is confict

let me fix that also

romani:
@MonuChaudhary14 , please resolve conflict

MonuChaudhary14:
@rdiachenko  @romani @sabaka 

Please check this PR

MonuChaudhary14:
@romani @sabaka @rdiachenko 

Ping

src/main/java/org/checkstyle/autofix/recipe/FinalClass.java
```diff
@@ -80,7 +65,15 @@ public J.ClassDeclaration visitClassDeclaration(
...
                     super.visitClassDeclaration(classDeclaration, executionContext);
 
-            if (isAtViolationLocation(classDeclaration) && !hasFinalModifier(result)) {
+            final boolean hasMarker = result.getMarkers()
+                    .findAll(CheckstyleViolationMarker.class)
+                    .stream()
+                    .map(CheckstyleViolationMarker::getViolation)
+                    .map(CheckstyleViolation::getSource)
+                    .map(CheckstyleCheck::checkName)
+                    .anyMatch(CheckFullName.FINAL_CLASS::equals);
```
rdiachenko: This logic will be duplicated across other recipes. Let's move it into `CheckstyleViolationMarker` to have something like:
```java
public boolean isFor(CheckFullName check) {
        return violation.getSource().checkName().equals(check);
    }
```
this also allows to get rid of exposed `getViolation` method

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
```diff
@@ -216,6 +137,14 @@ else if (nonViolations.isEmpty()) {
             }
         }
 
+        private boolean isViolationMarked(J.VariableDeclarations.NamedVariable variable) {
+            return variable.getMarkers().findAll(CheckstyleViolationMarker.class).stream()
+                    .map(CheckstyleViolationMarker::getViolation)
+                    .map(CheckstyleViolation::getSource)
+                    .map(CheckstyleCheck::checkName)
+                    .anyMatch(CheckFullName.FINAL_LOCAL_VARIABLE::equals);
```
rdiachenko: see comment above

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
```diff
@@ -53,15 +56,19 @@ public final class CheckstyleRecipeRegistry {
...
+            RECIPE_MAP_NO_VIOLATIONS = new EnumMap<>(CheckFullName.class);
+
     static {
         RECIPE_MAP.put(CheckFullName.AVOID_STAR_IMPORT, AvoidStarImport::new);
         RECIPE_MAP.put(CheckFullName.EMPTY_STATEMENT, EmptyStatement::new);
+        RECIPE_MAP_NO_VIOLATIONS.put(CheckFullName.FINAL_CLASS, FinalClass::new);
         RECIPE_MAP.put(CheckFullName.CONSTRUCTORS_DECLARATION_GROUPING,
             ConstructorsDeclarationGrouping::new);
-        RECIPE_MAP.put(CheckFullName.FINAL_CLASS, FinalClass::new);
+        RECIPE_MAP_NO_VIOLATIONS.put(CheckFullName.FINAL_CLASS, FinalClass::new);
```
rdiachenko: you already added it 2 lines above

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
```diff
@@ -53,15 +56,19 @@ public final class CheckstyleRecipeRegistry {
...
         RECIPE_MAP.put(CheckFullName.UPPER_ELL, UpperEll::new);
         RECIPE_MAP.put(CheckFullName.HEX_LITERAL_CASE, HexLiteralCase::new);
-        RECIPE_MAP.put(CheckFullName.FINAL_LOCAL_VARIABLE, FinalLocalVariable::new);
+        RECIPE_MAP_NO_VIOLATIONS.put(CheckFullName.FINAL_LOCAL_VARIABLE, FinalLocalVariable::new);
```
rdiachenko: Let's put all `RECIPE_MAP_NO_VIOLATIONS` as a separate group after `RECIPE_MAP`.

Could you create an issue to migrate the rest recipes to marker-based approach. Just list them as subtasks (one per recipe) in the issue description.

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -0,0 +1,159 @@
...
+import org.checkstyle.autofix.parser.CheckstyleViolation;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.marker.Marker;
+
+public class ViolationMarkerRecipe extends Recipe {
```
rdiachenko: Direction is right (mark once, fix by marker), but there are few problems:

1. O(N^2) per file. `preVisit` runs on every node, and `PositionHelper.computePosition` re-prints the whole tree each call to count `\n` and columns. This works bad for big files.
2. Strict (line, column) equality drops violations silently. Many checks point at positions no LST node starts at (e.g., whitespace, operators inside expressions, commas). Those markers never attach.

Let's try a different approach. `ScanningRecipe` keyed on `Tree.getId()`:

1. Scan phase walks each `J.CompilationUnit` once, finds the smallest LST node whose range contains each violation, stores `Map<UUID, CheckstyleViolationMarker>` in the accumulator. Expensive position math runs once per file, not per node per cycle.
2. Visit phase walks the tree, looks up by `tree.getId()`, attaches markers. O(N). `Tree.getId()` is stable across `withX(...)`, so markers survive every downstream fix and no `MarkersApplied` needed.

Here's a scatch that might help:
```java
  public class ViolationMarkerRecipe extends ScanningRecipe<ViolationMarkerRecipe.Accumulator> {

      private final List<CheckstyleViolation> violations;

      public ViolationMarkerRecipe(List<CheckstyleViolation> violations) {
          this.violations = List.copyOf(violations);
      }

      @Override
      public Accumulator getInitialValue(ExecutionContext ctx) { /* ... */ }

      // Read-only. For each CU, find the smallest LST node whose range contains
      // each violation, store by tree.getId() in acc. Position math lives here, once.
      @Override
      public TreeVisitor<?, ExecutionContext> getScanner(Accumulator acc) { /* ... */ }

      // Walk again, look up by tree.getId(), attach markers. No position math.
      @Override
      public TreeVisitor<?, ExecutionContext> getVisitor(Accumulator acc) { /* ... */ }

      static final class Accumulator {
          final Map<Path, Map<UUID, List<CheckstyleViolationMarker>>> byFile = new HashMap<>();
      }
  }
```

Hints:
- "Smallest containing range" = print the tree once, record (startLine, startCol)–(endLine, endCol) per node, pick the deepest node whose range contains the violation position. Keep this in a private helper.

rdiachenko (changes requested):

MonuChaudhary14:
@rdiachenko 

Please take a look at this approach, if it is correct, then i will try kill the mutations

rdiachenko:
> @rdiachenko
> 
> Please take a look at this approach, if it is correct, then i will try kill the mutations

Yep, the direction is right. Few things to address:

1. Fix the matcher, not the fixers.

The smallest enclosing range often picks `J.Modifier` or `J.Identifier` instead of the actual container the fixer cares about. That is why `FinalClass` has to also scan `result.getModifiers()` and `FinalLocalVariable.isViolationMarked` has to fall back to `variable.getName().getMarkers()`. Both of those are workarounds.

Direction: in `ScannerVisitor`, after you find the smallest enclosing cursor, walk up to the container type the check actually targets (`J.ClassDeclaration` for FINAL_CLASS, `J.VariableDeclarations.NamedVariable` for FINAL_LOCAL_VARIABLE), and attach the marker on that node instead. Use a small switch over `CheckFullName` for now. Once that's done, delete the parent/child fallback in both fixers, leaving just `findFirst(CheckstyleViolationMarker.class)` plus `isFor(...)`.

Open a follow-up issue: "Move violation target resolution into CheckFullName enum (predicate per check)". We will move the switch logic into the enum once a few more recipes are migrated.

2. Drop the System.identityHashCode UUID hack.

It happens to work because MarkersApplied already prevents the visitor from running twice, so the marker id never needs to be stable across cycles. Please recheck this, and if it's the case replace it with Tree.randomId() at the construction site. Also collapse MarkersApplied.withId to a single return statement.

3. Think how to avoid this magic constant `LINE_MULTIPLIER = 10000`

MonuChaudhary14:
@rdiachenko 

Please review this

src/main/java/org/checkstyle/autofix/recipe/FinalClass.java
```diff
@@ -57,20 +50,10 @@ public String getDescription() {
 
     @Override
     public TreeVisitor<?, ExecutionContext> getVisitor() {
-        return new FinalClassVisitor();
+        return new FixerVisitor();
```
rdiachenko: what's the reason for renaming? please revert

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
```diff
@@ -43,10 +39,7 @@
  */
 public class FinalLocalVariable extends Recipe {
 
-    private final List<CheckstyleViolation> violations;
-
-    public FinalLocalVariable(List<CheckstyleViolation> violations) {
-        this.violations = violations;
+    public FinalLocalVariable() {
```
rdiachenko: this is redundant, please remove

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
```diff
@@ -357,7 +359,43 @@ public J preVisit(J tree, ExecutionContext executionContext) {
             }
 
         }, new InMemoryExecutionContext(throwable -> {
-            throwable.printStackTrace();
+            if (throwable instanceof RuntimeException) {
+                throw (RuntimeException) throwable;
+            }
+            throw new RuntimeException(throwable);
```
rdiachenko: this looks like unexpected behaviour, let's throw IllegalStateException(throwable), and remove check for instancof, should be a single line throw

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -0,0 +1,404 @@
...
+            }
+
+            if (!markersToAdd.isEmpty()) {
+                acc.putByFile(sourcePath, markersToAdd);
+            }
+        }
+
+        private void findSmallestNode(CheckstyleViolation violation, Map<UUID, Range> nodeRanges,
+                                      Map<UUID, UUID> parentMap, Map<UUID, Tree> nodeTrees,
+                                      Map<UUID, List<CheckstyleViolationMarker>> markersToAdd) {
```
rdiachenko: findSmallestNode mutates markersToAdd passed in. Returning the result instead would be cleaner

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -0,0 +1,404 @@
...
+            }
+            return result;
+        }
+    }
+
+    private static final class BoundingBoxCapture extends PrintOutputCapture<TreeVisitor<?, ?>> {
+        private final Map<UUID, Range> nodeRanges;
+        private final Map<UUID, UUID> parentMap;
+        private final Map<UUID, Tree> nodeTrees;
+        private int line = 1;
```
rdiachenko: move initialization to constructor

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -0,0 +1,404 @@
...
+            this.line = line;
+        }
+
+        public int getColumn() {
+            return column;
+        }
+
+        public void setColumn(int column) {
+            this.column = column;
+        }
```
rdiachenko: no need for these getters/setters since they are used within this class only, use direct fields instead

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -0,0 +1,404 @@
...
+                    for (CheckstyleViolationMarker marker : markers) {
+                        result = result.withMarkers(result.getMarkers().add(marker));
+                    }
+                }
+            }
+            return result;
+        }
+    }
+
+    public static final class Accumulator {
```
rdiachenko: let's move it into its own file

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -0,0 +1,404 @@
...
+                }
+                else {
+                    setColumn(getColumn() + 1);
+                }
+            }
+
+            for (Tree treeNode : pathTrees) {
+                final Range range = nodeRanges.get(treeNode.getId());
+                range.setEndLine(getLine());
+                range.setEndCol(getColumn());
```
rdiachenko: this is not good, you get an object from the map and modifies it, which can lead to hard-to-debug bugs, since the map now contains modified object and other places that use this map may not expect this.

Make `Range` an immutable record, it shouldn't contain any setters. And update the rest logic accordingly.

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -0,0 +1,404 @@
...
+        public int getEndCol() {
+            return endCol;
+        }
+
+        public void setEndCol(int endCol) {
+            this.endCol = endCol;
+        }
+    }
+
+    record MarkersApplied(UUID id) implements Marker {
```
rdiachenko: make it private

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -0,0 +1,404 @@
...
+        @Override
+        @SuppressWarnings("unchecked")
+        public <M extends Marker> M withId(UUID uuid) {
+            final M result;
+            if (uuid.equals(id)) {
+                result = (M) this;
+            }
+            else {
+                result = (M) new MarkersApplied(uuid);
+            }
```
rdiachenko: records are immutable, the equality branch returns the same logical value either way. Collapse to one line: `return (M) new MarkersApplied(uuid)`

rdiachenko (commented):
Notes below

MonuChaudhary14:
@rdiachenko 

I have made the changes, please check these

src/main/java/org/checkstyle/autofix/recipe/FinalClass.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/FinalLocalVariable.java
MonuChaudhary14: done

src/test/java/org/checkstyle/autofix/recipe/AbstractRecipeTestSupport.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

MonuChaudhary14:
> > @rdiachenko
> > Please take a look at this approach, if it is correct, then i will try kill the mutations
> 
> Yep, the direction is right. Few things to address:
> 
> 1. Fix the matcher, not the fixers.
> 
> The smallest enclosing range often picks `J.Modifier` or `J.Identifier` instead of the actual container the fixer cares about. That is why `FinalClass` has to also scan `result.getModifiers()` and `FinalLocalVariable.isViolationMarked` has to fall back to `variable.getName().getMarkers()`. Both of those are workarounds.
> 
> Direction: in `ScannerVisitor`, after you find the smallest enclosing cursor, walk up to the container type the check actually targets (`J.ClassDeclaration` for FINAL_CLASS, `J.VariableDeclarations.NamedVariable` for FINAL_LOCAL_VARIABLE), and attach the marker on that node instead. Use a small switch over `CheckFullName` for now. Once that's done, delete the parent/child fallback in both fixers, leaving just `findFirst(CheckstyleViolationMarker.class)` plus `isFor(...)`.
> 
> Open a follow-up issue: "Move violation target resolution into CheckFullName enum (predicate per check)". We will move the switch logic into the enum once a few more recipes are migrated.
> 
> 2. Drop the System.identityHashCode UUID hack.
> 
> It happens to work because MarkersApplied already prevents the visitor from running twice, so the marker id never needs to be stable across cycles. Please recheck this, and if it's the case replace it with Tree.randomId() at the construction site. Also collapse MarkersApplied.withId to a single return statement.
> 
> 3. Think how to avoid this magic constant `LINE_MULTIPLIER = 10000`



done, and here is the link to issue

https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/231

src/main/java/org/checkstyle/autofix/CheckstyleRecipeRegistry.java
MonuChaudhary14: done 

issue to link
https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/234

rdiachenko (commented):
Good as the first version. There are some corner cases (e.g. column number can be absent, so we use -1 as default value) but we'll catch those when we get into such cases.

@MonuChaudhary14 please make ci green and let's merge it. We'll discover issues and improve the logic when migrating other recipes.

Please also create an issue per recipe to migrate all current recipes to marker-based approach.

MonuChaudhary14:
@rdiachenko 

I have some doubt regarding killing these two mutations

```
+    <sourceFile>FinalClass.java</sourceFile>
+    <mutatedClass>org.checkstyle.autofix.recipe.FinalClass$FinalClassVisitor</mutatedClass>
+    <mutatedMethod>lambda$visitClassDeclaration$0</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.BooleanTrueReturnValsMutator</mutator>
+    <description>replaced boolean return with true for org/checkstyle/autofix/recipe/FinalClass$FinalClassVisitor::lambda$visitClassDeclaration$0</description>
+    <lineContent>.anyMatch(marker -&gt; marker.isFor(CheckFullName.FINAL_CLASS));</lineContent>
+  </mutation>
```

```
+    <sourceFile>FinalLocalVariable.java</sourceFile>
+    <mutatedClass>org.checkstyle.autofix.recipe.FinalLocalVariable$LocalVariableVisitor</mutatedClass>
+    <mutatedMethod>lambda$isViolationMarked$0</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.BooleanTrueReturnValsMutator</mutator>
+    <description>replaced boolean return with true for org/checkstyle/autofix/recipe/FinalLocalVariable$LocalVariableVisitor::lambda$isViolationMarked$0</description>
+    <lineContent>.anyMatch(marker -&gt; marker.isFor(CheckFullName.FINAL_LOCAL_VARIABLE));</lineContent>
+  </mutation>
```

These mutations are from these lines 

```
final boolean hasMarker = result.getMarkers()
                    .findAll(CheckstyleViolationMarker.class).stream()
                    .anyMatch(marker -> marker.isFor(CheckFullName.FINAL_CLASS));
                
```

PIT targeted the lambda function `marker -> marker.isFor(CheckFullName.FINAL_CLASS)`. Because this lambda returns a boolean, PIT applied the `BooleanTrueReturnValsMutator`, which modifies the compiled code to ignore whatever the lambda was supposed to do and simply make it: marker -> true

Case 1: A class with no violation marker-> When the code scans the class, teh stream is completely empty. 
`anyMatch` ona empty stream always returns false immediately, so the lambda is never even called. The class is correctly left alone, and the test passes.

Case 2: A class with violation -> When the code scans the class, the stream has one marker. 
`anyMatch` calls the mutated lambda, which returns true. The class is modified, and the test passes.
Even through it blindly returns the true, the test passes.

The scenerio, i can think is that, the class contain the violation other than recipe violation,  where it will have a different checkstyle marker

IN original code, the lambda will return a false, and the class would not be modified, but in mutated code, the lambda would return true, and falsy assume that it is recipe violtion, and the class would be wrongly modify, and the test will fail

We cannot add such types of test, because we cannot have checkstyle violations on Output files, so we cannot create these types of test


There are other two methods, which we can do

1. Instead of using `.stream().anyMatch(lambda)` we use a classic for loop inside a static helper method.

In `CheckstyleViolationMarker.java`:
```
public static boolean hasViolation(Markers markers, CheckFullName check) {
    for (CheckstyleViolationMarker marker : markers.findAll(CheckstyleViolationMarker.class)) {
        if (marker.isFor(check)) {
            return true;
        }
    }
    return false;
}
```

In recipes(Finalclass, FinalLocalVariable):
```
final boolean hasMarker = CheckstyleViolationMarker.hasViolation(result.getMarkers(), CheckFullName.FINAL_CLASS);
```

By moving to a for loop, the lambda `marker -> marker.isFor(...)` is completely removed from codebase. With no lambda returning a boolean, PIT has absolutely nothing to apply the BooleanTrueReturnValsMutator to. ALso, this perfectly satisfies the deduplicating the logic and keeping .violation() out of the recipes.

2. We can map the stream all the way to the CheckFullName enum and use the built-in contains or method reference.
In recipe:
```
final boolean hasMarker = result.getMarkers()
        .findAll(CheckstyleViolationMarker.class).stream()
        .map(CheckstyleViolationMarker::violation)
        .map(CheckstyleViolation::getSource)
        .map(CheckstyleCheck::checkName)
        .toList()
        .contains(CheckFullName.FINAL_CLASS);
```

contains() doesn't use a lambda predicate, so PIT cannot mutate it to return true

Also, should I suppress the remaining mutations of ViolationMarkerRecipe and CheckstyleViolationMarker? Since they are mostly boundary logic and interface methods, they will require pure unit tests to kill them.

rdiachenko:
1. The for-loop does remove the lambda, but PIT still has equivalent mutators that hit the new code. Please recheck and confirm. Better to suppress this specific case with the comment "Cannot be killed by integration tests until a second recipe lands a marker on this node type. See issue #XXX". Open a follow-up issue to revisit once all recipes are migrated.

2. No tests is worse than thin tests. Please cover the logic with unit tests now, no mutations suppressed here. Open a follow-up issue to revisit after all recipes are migrated to the marker-based approach. By then many of these mutations will be killed by other recipes tests, and we can drop the unit tests that became redundant.

romani:
Yes, we will be back to pitest cleanup, and we will have more test coverage from real recipes

MonuChaudhary14:
@rdiachenko 

Please check the unit tests

These are the two issues, that you have asked for
https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/236
https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/237

There are 4 TIMED_OUT mutations in this, these mutations require large changes in ViolationMarkerRecipe

rdiachenko (approved):
Ok to merge after this point is addressed from [this comment](https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/153#pullrequestreview-4461872435):

> Please also create an issue per recipe to migrate all current recipes to marker-based approach.

MonuChaudhary14:
> Ok to merge after this point is addressed from [this comment](https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/153#pullrequestreview-4461872435):
> 
> > Please also create an issue per recipe to migrate all current recipes to marker-based approach.

Please check this
Issue link: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/234

Issue #154, opened 2026-03-01
implement recipe UseEnhancedSwitch


romani:
from https://github.com/checkstyle/checkstyle/issues/15180 , released in 13.3.0

it can be automated, 
see example of fixes by Check https://github.com/camunda/camunda/pull/46417

romani:
@MonuChaudhary14 , can you help world to upgrade to better coding style in automated way ?

MonuChaudhary14:
> [@MonuChaudhary14](https://github.com/MonuChaudhary14) , can you help world to upgrade to better coding style in automated way ?

Thank you for the opportunity
Yes, I would be happy to contribute on the UseEnhancedSwitch recipe

romani:
Merged

PR #155, opened 2026-03-02
Issue #154: Implemented the recipe for UseEnhancedSwitch

MonuChaudhary14:
Issue #154

Implemented the recipe for UseEnhancedSwitch

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -45,7 +45,7 @@
 
         <!-- Checkstyle properties -->
         <maven.checkstyle.plugin.version>3.6.0</maven.checkstyle.plugin.version>
-        <checkstyle.version>13.2.0</checkstyle.version>
+        <checkstyle.version>13.3.0</checkstyle.version>
 
         <!-- Pitest properties -->
         <pitest.version>1.23.1</pitest.version>
--- a/src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
+++ b/src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
@@ -1,0 +1,712 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle-openrewrite-recipes: Automatically fix Checkstyle violations with OpenRewrite.
+// Copyright (C) 2025 The Checkstyle OpenRewrite Recipes Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package org.checkstyle.autofix.recipe;
+
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.checkstyle.autofix.PositionHelper;
+import org.checkstyle.autofix.parser.CheckstyleViolation;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.internal.RecipeRunException;
+import org.openrewrite.java.JavaVisitor;
+import org.openrewrite.java.tree.Expression;
+import org.openrewrite.java.tree.Flag;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JContainer;
+import org.openrewrite.java.tree.JLeftPadded;
+import org.openrewrite.java.tree.JavaType;
+import org.openrewrite.java.tree.JavaType.FullyQualified;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.Statement;
+import org.openrewrite.marker.Markers;
+
+/**
+ * Fixes Checkstyle UseEnhancedSwitch violations by converting traditional switch
+ * statements using colon syntax to enhanced switch using arrow syntax.
+ */
+public class UseEnhancedSwitch extends Recipe {
+
+    private final List<CheckstyleViolation> violations;
+
+    public UseEnhancedSwitch(List<CheckstyleViolation> violations) {
+        this.violations = violations;
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "UseEnhancedSwitch recipe";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Convert switch statements using colon syntax to enhanced switch "
+                + "using arrow syntax.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new UseEnhancedSwitchVisitor();
+    }
+
+    private final class UseEnhancedSwitchVisitor extends JavaVisitor<ExecutionContext> {
+
+        private static final String DEFAULT_LABEL = "default";
+        private Path sourcePath;
+        private J.CompilationUnit compUnit;
+
+        @Override
+        public J.CompilationUnit visitCompilationUnit(
+                J.CompilationUnit cu, ExecutionContext executionContext) {
+            this.sourcePath = cu.getSourcePath();
+            this.compUnit = cu;
+            return (J.CompilationUnit) super.visitCompilationUnit(cu, executionContext);
+        }
+
+        @Override
+        public Statement visitSwitch(J.Switch switchNode, ExecutionContext executionContext) {
+            final J.Switch visited = (J.Switch) super.visitSwitch(switchNode, executionContext);
+            Statement result = visited;
+
+            if (isTraditional(visited.getCases().getStatements())
+                    && isAtViolationLocation(visited)) {
+                final List<Statement> mergedCases = combineCasesWithSameStatements(
+                        visited.getCases().getStatements());
+                final boolean isExpr = canConvertToReturnSwitchExpression(
+                        mergedCases, isEnumExhaustive(visited.getSelector(), mergedCases));
+
+                final J.Switch updatedSwitch = visited.withCases(
+                        convertCases(visited.getCases(), visited.getSelector(), isExpr));
+
+                if (isExpr) {
+                    final J.SwitchExpression switchExpr = new J.SwitchExpression(
+                            Tree.randomId(),
+                            Space.SINGLE_SPACE,
+                            Markers.EMPTY,
+                            updatedSwitch.getSelector(),
+                            updatedSwitch.getCases(),
+                            null);
+                    result = new J.Return(
+                            Tree.randomId(),
+                            updatedSwitch.getPrefix(),
+                            Markers.EMPTY,
+                            switchExpr);
+                }
+                else {
+                    result = updatedSwitch;
+                }
+            }
+            return result;
+        }
+
+        private boolean isTraditional(final List<? extends Statement> cases) {
+            return !cases.isEmpty() && ((J.Case) cases.get(0)).getType() == J.Case.Type.Statement;
+        }
+
+        @Override
+        public J.Block visitBlock(J.Block block, ExecutionContext executionContext) {
+            final J.Block visited = (J.Block) super.visitBlock(block, executionContext);
+            return mergeAssignmentSwitches(visited);
+        }
+
+        private J.Block removeUnreachableCode(final J.Block visited) {
+            final List<Statement> stmts = visited.getStatements();
+            final List<Statement> newStmts = new ArrayList<>();
+            for (final Statement s : stmts) {
+                newStmts.add(s);
+                if (s instanceof J.Return) {
+                    break;
+                }
+            }
+            return visited.withStatements(newStmts);
+        }
+
+        private J.Block mergeAssignmentSwitches(final J.Block visited) {
+            final List<Statement> stmts = visited.getStatements();
+            final List<Statement> merged = new ArrayList<>();
+
+            int index = 0;
+            while (index < stmts.size()) {
+                int jump = 1;
+                if (index < stmts.size() - 1) {
+                    final Statement s1 = stmts.get(index);
+                    final Statement s2 = stmts.get(index + 1);
+                    if (s1 instanceof J.VariableDeclarations vd
+                            && vd.getVariables().size() == 1
+                            && vd.getVariables().get(0).getInitializer() == null
+                            && s2 instanceof J.Switch sw) {
+                        final String name = vd.getVariables().get(0)
+                                .getName().getSimpleName();
+                        final List<Statement> mergedCases =
+                                sw.getCases().getStatements();
+                        final boolean exhaustive =
+                                isEnumExhaustive(sw.getSelector(), mergedCases);
+                        if (canConvertToAssignmentSwitchExpression(
+                                mergedCases, name, exhaustive)) {
+                            final J.VariableDeclarations converted =
+                                    convertToAssignmentSwitchExpression(
+                                            vd, sw, name);
+                            merged.add(converted);
+                            jump = 2;
+                        }
+                    }
+                }
+                if (jump == 1) {
+                    merged.add(stmts.get(index));
+                }
+                index += jump;
+            }
+            return removeUnreachableCode(visited.withStatements(merged));
+        }
+
+        private J.VariableDeclarations convertToAssignmentSwitchExpression(
+                J.VariableDeclarations varDecl, J.Switch switchStmt, String varName) {
+            final J.Block newCases = convertAssignmentCases(switchStmt.getCases(), varName);
+
+            final J.SwitchExpression switchExpr = new J.SwitchExpression(
+                    Tree.randomId(),
+                    Space.SINGLE_SPACE,
+                    Markers.EMPTY,
+                    switchStmt.getSelector(),
+                    newCases,
+                    null);
+
+            J.VariableDeclarations.NamedVariable namedVar = varDecl.getVariables().get(0);
+            final JLeftPadded<Expression> initializer = JLeftPadded
+                    .build((Expression) switchExpr)
+                    .withBefore(Space.SINGLE_SPACE);
+            namedVar = namedVar.getPadding().withInitializer(initializer);
+            final List<J.VariableDeclarations.NamedVariable> variables =
+                    new ArrayList<>(varDecl.getVariables());
+            variables.set(0, namedVar);
+            return varDecl.withVariables(variables);
+        }
+
+        private J.Block convertAssignmentCases(final J.Block casesBlock, final String varName) {
+            final List<Statement> caseStatements = casesBlock.getStatements();
+            final List<Statement> newStatements = new ArrayList<>();
+            for (final Statement stmt : caseStatements) {
+                final J.Case converted = convertAssignmentCase((J.Case) stmt, varName);
+                newStatements.add(converted);
+            }
+            return casesBlock.withStatements(newStatements);
+        }
+
+        private J.Case convertAssignmentCase(J.Case caseStmt, String varName) {
+            return convertArrowAssignmentCase(caseStmt, varName);
+        }
+
+        private J.Case convertArrowAssignmentCase(final J.Case caseStmt, final String varName) {
+            final J body = caseStmt.getBody();
+            final J.Case result;
+            if (body instanceof J.Block block) {
+                final List<Statement> adjusted =
+                        adjustForAssignment(block.getStatements(), varName);
+                result = caseStmt.withBody(block.withStatements(adjusted));
+            }
+            else {
+                final J newBody;
+                if (body instanceof J.Assignment asgn) {
+                    newBody = asgn.getAssignment();
+                }
+                else {
+                    newBody = body;
+                }
+                final J prefixed = newBody;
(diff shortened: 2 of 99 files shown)


MonuChaudhary14:
@romani @rdiachenko @Anmol202005 

Please review this PR

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,212 @@
...
+                    PositionHelper.computeColumnPosition(cu, switchStatement, getCursor());
+
+            return violations.stream().anyMatch(violation -> {
+                return violation.getLine() == line
+                        && violation.getColumn() == column
+                        && violation.getFilePath().endsWith(sourcePath);
+            });
+        }
+
+        private boolean isAtViolationLocationForExpression(J.SwitchExpression switchExpression) {
```
Anmol202005: `isAtViolationLocation` and `isAtViolationLocationForExpression` are same.
these can be merged into a single method:
```
  private boolean isAtViolationLocation(J switchNode) {
      
  }
```

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,212 @@
...
+                    .withPrefix(Space.SINGLE_SPACE)
+                    .withStatements(new ArrayList<>(statements))
+                    .withEnd(computeBlockEndSpace(statements)));
+        }
+
+        private J.Case setArrowType(J.Case caseStmt) {
+            return caseStmt.withType(J.Case.Type.Rule)
+                    .getPadding().withStatements(JContainer.build(
+                            Space.SINGLE_SPACE, new java.util.ArrayList<>(),
+                            org.openrewrite.marker.Markers.EMPTY));
```
Anmol202005: Markers should be added as an import

Anmol202005 (changes requested):
changes:

Anmol202005:
@MonuChaudhary14 Also update the README

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: done

MonuChaudhary14:
> @MonuChaudhary14 Also update the README

done

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new UseEnhancedSwitchVisitor();
+    }
+
+    private final class UseEnhancedSwitchVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        private static final int INDENT_SIZE = 4;
```
Anmol202005: Please avoid hardcoding this value. Instead, retrieve the indent value directly from the configuration

Anmol202005 (changes requested):

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,210 @@
...
+                + "using arrow syntax.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new UseEnhancedSwitchVisitor(getIndentSize());
+    }
+
+    private int getIndentSize() {
+        return Integer.parseInt(config.getProperty("tabWidth"));
```
Anmol202005: why do we need to store the indent size?
Is it to figure out the position of the closing `}` of the generated block? If so, I think we don't need to compute that at all the `case` node itself already has its own prefix whitespace, which is exactly the indent we want for the closing brace.

This would let us remove `computeBlockEndSpace`, `computeIndent`, the `configuredIndentSize` field, the visitor constructor parameter, and `getIndentSize()` in the outer class.

romani:
@MonuChaudhary14 , please test your implementation on comunda 
Example https://github.com/camunda/camunda/pull/46417

Diff should be same , only spacing/indentation should a difference.

Anmol202005 (approved):
lgtm

romani:
@MonuChaudhary14 , please fix pitest, no mutations allowed

romani:
@MonuChaudhary14  , what is state of applying changes to comunda ?

MonuChaudhary14:
> @MonuChaudhary14 , what is state of applying changes to comunda ?


here is the diff report
https://gist.githubusercontent.com/MonuChaudhary14/9a6b8ba2f419d0e762698488be38bc97/raw/79048d60935fd71ce872010fd6f94d641490f731/enhancedSwitch.diff

I have made a lot of changes in the recipe to make the output similar to camunda and the recipe is too big, there is a lot of mutations, so i need some time to kill all the mutations

romani:
Nice report.
```

diff --git a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/entity/ProcessInstanceEntityTransformer.java b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/entity/ProcessInstanceEntityTransformer.java
index f41cdd1803c..4da4355b103 100644
--- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/entity/ProcessInstanceEntityTransformer.java
+++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/entity/ProcessInstanceEntityTransformer.java
@@ -46,15 +46,10 @@ public class ProcessInstanceEntityTransformer
     if (source == null) {
       return null;
     }
-    switch (source) {
-      case ACTIVE:
-        return ProcessInstanceState.ACTIVE;
-      case COMPLETED:
-        return ProcessInstanceState.COMPLETED;
-      case CANCELED:
-        return ProcessInstanceState.CANCELED;
-      default:
-        throw new IllegalArgumentException("Unknown process instance state: " + source);
-    }
+    return switch (source) {
+      case ACTIVE -> ProcessInstanceState.ACTIVE;
+      case COMPLETED -> ProcessInstanceState.COMPLETED;
+      case CANCELED -> ProcessInstanceState.CANCELED;
+    };
   }
 }
```

Default is missing in resulted code

MonuChaudhary14:
> Nice report.
> 
> ```
> 
> diff --git a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/entity/ProcessInstanceEntityTransformer.java b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/entity/ProcessInstanceEntityTransformer.java
> index f41cdd1803c..4da4355b103 100644
> --- a/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/entity/ProcessInstanceEntityTransformer.java
> +++ b/search/search-client-query-transformer/src/main/java/io/camunda/search/clients/transformers/entity/ProcessInstanceEntityTransformer.java
> @@ -46,15 +46,10 @@ public class ProcessInstanceEntityTransformer
>      if (source == null) {
>        return null;
>      }
> -    switch (source) {
> -      case ACTIVE:
> -        return ProcessInstanceState.ACTIVE;
> -      case COMPLETED:
> -        return ProcessInstanceState.COMPLETED;
> -      case CANCELED:
> -        return ProcessInstanceState.CANCELED;
> -      default:
> -        throw new IllegalArgumentException("Unknown process instance state: " + source);
> -    }
> +    return switch (source) {
> +      case ACTIVE -> ProcessInstanceState.ACTIVE;
> +      case COMPLETED -> ProcessInstanceState.COMPLETED;
> +      case CANCELED -> ProcessInstanceState.CANCELED;
> +    };
>    }
>  }
> ```
> 
> Default is missing in resulted code

In this case, source is an enum and all enum constants are handled in the switch expression, so it is exhaustive and a default branch is not required.
If a new enum constant is introduced in the future, the compiler will fail, forcing the switch to be updated

This is also done in camunda
https://github.com/camunda/camunda/pull/46417/changes#diff-9849fa92ffd76c1cb9647489581a8b03481149a340df7a43a970b51dda068eb7L814

https://github.com/camunda/camunda/pull/46417/changes#diff-dc61c1308702ba5ff9656038d66d1e3b145245614de441abf05122f3bb6a1729L31

https://github.com/camunda/camunda/pull/46417/changes#diff-362e5a7cc57c51c86d2814847df71e6b67c21b1cbe7a39c3d268896c10392200L42

romani:
@MonuChaudhary14 , please look at https://github.com/checkstyle/checkstyle/pull/19114#issuecomment-4024495701
What is your opinion on openrewrite attempt to do this?
I hope your implementation can avoid such problems or we can notice a gap in our implementation

MonuChaudhary14:
> @MonuChaudhary14 , please look at [checkstyle/checkstyle#19114 (comment)](https://github.com/checkstyle/checkstyle/pull/19114#issuecomment-4024495701) What is your opinion on openrewrite attempt to do this? I hope your implementation can avoid such problems or we can notice a gap in our implementation

Could you please clarify which issue you are referring to in that comment? I would like to better understand the concern so I can check whether our implementation might run into the same problem.

romani:
My concern is cleared. Check does not do violation, so no auto fix is possible.

Please fix pitest, no survival are allowed

MonuChaudhary14:
> My concern is cleared. Check does not do violation, so no auto fix is possible.
> 
> Please fix pitest, no survival are allowed

I’ve been working on it-was a bit busy earlier. I’ve already killed 150+ mutations  and only a few are left now

romani:
keep working , a pack of mutations still left.

if you run into lack of ideas how to kill: apply what pitest is suggesting and run on some big code base to see patch.

MonuChaudhary14:
> keep working , a pack of mutations still left.
> 
> if you run into lack of ideas how to kill: apply what pitest is suggesting and run on some big code base to see patch.

Thank you! I’m currently stuck on a few remaining mutations, so I’ll follow your suggestion and work on addressing them as soon as possible.

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,706 @@
...
+                if (!(stmt instanceof J.Break breakStmt)
+                        || breakStmt.getLabel() != null) {
+                    filtered.add(stmt);
+                }
+            }
+            return filtered;
+        }
+
+        private boolean isAtViolationLocation(final J switchNode) {
+            final J.CompilationUnit cu = getCursor().firstEnclosing(J.CompilationUnit.class);
```
romani: Let's use at least 3 characters for names.
Not sure why checkstyle is not barking on this.

This can be `unit` , or `compUnit`.

Please fix all other two chars variables in this PR

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,706 @@
...
+        }
+
+        private static final class ControlFlowVisitor extends JavaVisitor<Void> {
+            @Override
+            public J.Return visitReturn(J.Return return_, Void v) {
+                throw new ControlFlowException();
+            }
+
+            @Override
+            public J.Break visitBreak(J.Break break_, Void v) {
```
romani: You likely took this naming from method of library.
Let's not polute our coding with foreign naming convention.

`visitBreak(J.Break breakObj, Void voidObj) {`

romani (changes requested):
Items

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: done

MonuChaudhary14:
@romani 

Can you please review this

MonuChaudhary14:
@sabaka 
Please review this PR

pom.xml
```diff
@@ -337,10 +337,9 @@
                                 <outputFormat>HTML</outputFormat>
                             </outputFormats>
 
-                            <threads>4</threads>
-                            <timeoutFactor>10</timeoutFactor>
-                            <timeoutConstant>50000</timeoutConstant>
-
+                            <timeoutFactor>100</timeoutFactor>
+                            <timeoutConstant>100000</timeoutConstant>
+                            <jvmArgs>-Xmx8g</jvmArgs>
```
sabaka: What is the reason for this change?

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,706 @@
...
+            final List<Statement> stmts = cases.getStatements();
+            return !stmts.isEmpty() && ((J.Case) stmts.get(0)).getType() == J.Case.Type.Statement;
+        }
+
+        @Override
+        public Statement visitSwitch(J.Switch switchNode, ExecutionContext executionContext) {
+            final J.Switch visited = (J.Switch) super.visitSwitch(switchNode, executionContext);
+            Statement result = visited;
+
+            if (isTraditional(visited.getCases()) && isAtViolationLocation(visited)) {
```
sabaka: Is there a possible case when `isAtViolationLocation` would return `true` but `isTraditional` returns false? If violation is only triggered on traditional switches, what is the reason for checking for switch type?

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,706 @@
...
+
+    private final class UseEnhancedSwitchVisitor extends JavaVisitor<ExecutionContext> {
+
+        private static final String DEFAULT_LABEL = "default";
+        private Path sourcePath;
+
+        @Override
+        public J.CompilationUnit visitCompilationUnit(
+                J.CompilationUnit compUnit, ExecutionContext executionContext) {
+            this.sourcePath = compUnit.getSourcePath();
```
sabaka: It feels like saving compilation unit is more effective. You use both In `isAtViolationLocation` - now you extract compilation unit from a switch node which (i think) walks a tree every time. If you store CU in a field and access sourcePath in `isAtViolationLocation` it would be just field access.

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,706 @@
...
+            return (J.CompilationUnit) super.visitCompilationUnit(compUnit, executionContext);
+        }
+
+        private boolean isTraditional(final J.Block cases) {
```
sabaka: move to be after visitSwitch

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,706 @@
...
+
+        private boolean isEnumExhaustive(
+                final J.ControlParentheses<Expression> selector,
+                final List<Statement> mergedCases) {
+            boolean result = false;
+            final JavaType type = selector.getTree().getType();
+            if (type instanceof JavaType.FullyQualified fq
+                    && fq.getKind() == JavaType.FullyQualified.Kind.Enum) {
+                final List<JavaType.Variable> members = fq.getMembers();
+                final java.util.Set<String> handled = collectHandled(mergedCases);
```
sabaka: Could be import

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,706 @@
...
+        @Override
+        public J.SwitchExpression visitSwitchExpression(
+                J.SwitchExpression switchExpression, ExecutionContext executionContext) {
+            final J.SwitchExpression visited = (J.SwitchExpression) super.visitSwitchExpression(
+                    switchExpression, executionContext);
+
+            J.SwitchExpression result = visited;
+            final boolean traditional = isTraditional(visited.getCases());
+            final boolean violation = isAtViolationLocation(visited);
+            if (traditional && violation) {
```
sabaka, line 266: Same question as above, can it be non-traditional but have violation?

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,706 @@
...
+            }
+
+            if (!pendingCases.isEmpty()) {
+                result.add(buildMergedCase(pendingCases));
+            }
+            return result;
+        }
+
+        private J.Case buildMergedCase(final List<J.Case> cases) {
+            final J.Case last = cases.get(cases.size() - 1);
```
sabaka: getLast()

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,706 @@
...
+                    handledConstants.add(((J.Identifier) label).getSimpleName());
+                }
+            }
+            return handledConstants;
+        }
+
+        private List<Statement> removeDefaultCase(final List<Statement> statements) {
+            return statements.stream()
+                    .filter(stmt -> !isDefaultCase((J.Case) stmt))
+                    .collect(java.util.stream.Collectors.toList());
```
sabaka: could be an import

sabaka: actually, can be just `.toList()`

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,706 @@
...
+        }
+
+        private List<Statement> mergeFallthroughCases(
```
sabaka: is naming correct? I don't see how fallthrough is handled here

pom.xml
MonuChaudhary14: i changed it during pitest testing, i will revert it again

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: > Is there a possible case when `isAtViolationLocation` would return `true` but `isTraditional` returns false? If violation is only triggered on traditional switches, what is the reason for checking for switch type?

This is a defensive case. When the violation list is created and the recipe starts executing, both conditions are true and the recipe applies the changes. However, when OpenRewrite runs the next cycle, it checks the violation list again (which is still there), but the isTraditional check now returns false because the switch has already been converted. This prevents an infinite loop of conversion. Furthermore, during testing on Camunda, we found that removing this guard causes a NoSuchElementException because the transformation logic assumes a traditional switch structure

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: i have changed the name

MonuChaudhary14:
@sabaka 
Please review this PR

MonuChaudhary14:

Please review this PR

romani (approved):
thanks a lot

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -0,0 +1,712 @@
...
+            for (Statement stmt : statements) {
+                if (!(stmt instanceof J.Break breakStmt)
+                        || breakStmt.getLabel() != null) {
+                    filtered.add(stmt);
+                }
+            }
+            return filtered;
+        }
+
+        private boolean isAtViolationLocation(final J switchNode) {
```
sabaka, line 624: I feel like we do that in every receipt. Could you please check if there is already an issue exists in the project with suggestion to extract it somewhere and create one if it doesn't.

sabaka (approved):

sabaka (approved):
Looks good, thank you

romani:
@MonuChaudhary14 , please rebase to resolve conflict

MonuChaudhary14:
> @MonuChaudhary14 , please rebase to resolve conflict

done

PR #157, opened 2026-03-07
infra: add editor config

yuluo-yx:

PR #172, opened 2026-04-18
infra: add editor config

yuluo-yx:
refer #157

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,45 @@
+# EditorConfig is awesome: https://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+
+max_line_length = 100
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.json]
+tab_width = 2
+indent_size = 2
+
+[*.{yml,yaml}]
+tab_width = 2
+indent_size = 2
+
+[*.xml]
+ij_xml_attribute_wrap = off
+ij_xml_text_wrap = off
+ij_xml_keep_blank_lines = 1
+
+[pom.xml]
+indent_size = 2
+
+[*.java]
+indent_size = 4
+tab_width = 4
+
+[*.md]
+trim_trailing_whitespace = false
+
+[*.properties]
+indent_size = 4
+
+[*.sh]
+indent_size = 2
+
+[Makefile]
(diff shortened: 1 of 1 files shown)


romani:
@MonuChaudhary14 , do you think it is valuable config ? does it works well with checkstyle config ?

MonuChaudhary14:
> @MonuChaudhary14 , do you think it is valuable config ? does it works well with checkstyle config ?

Let me look into this in detail and verify how well it aligns with the Checkstyle configuration. I’ll get back to you shortly

MonuChaudhary14:
> @MonuChaudhary14 , do you think it is valuable config ? does it works well with checkstyle config ?

I’ve checked out the config and I don't see any harm in using this. It’s actually quite helpful for catching those basic mistakes that new contributors often make, like forgetting the final newline or using incorrect indentation. Itbasically keeps the PRs cleaner from the start.

Technically, it works as a pre-commit formatting layer. It makes the IDE's auto-format feature line up with our standards, so contributors can fix those simple whitespace issues with a single shortcut before they even submit their code.

I even looked into the wiki (https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#ideas-for-domain-specific-properties) to see if there are newer domain-specific properties we could use. There are some ideas there, but since their main repo has been a bit inactive lately (last updated about a year ago), we really have to stick with the stable rules that are already supported by the format's Readers (like the IDE's formatting engine).

But yeah, it definitely has its limits.it can only toggle rules that are already built into the tool reading the file. It’s not a replacement for the deep, AST-based logic we have in Checkstyle.It's more of a proactive assistant for handling the low-hanging fruit automatically

romani:
@rdiachenko , please review and merge.

.editorconfig
```diff
@@ -0,0 +1,46 @@
...
+# top-most EditorConfig file
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_style = space
+
+# The line max length is 120, match checkstyle rules.
+max_line_length = 100
```
rdiachenko: comment says 120 but the real value is 100, where's the truth? please either remove the comment or update it to avoid mentioning any value

romani: Remove comment

yuluo-yx: done

romani:
Single commit please

rdiachenko (approved):
lgtm

Issue #198, opened 2026-05-23
Refactor usages of OpenRewrite internal API


MonuChaudhary14:
# Description

Currently, the project uses the internal OpenRewrite API org.openrewrite.internal.RecipeRunException in two files:

* src/main/java/org/checkstyle/autofix/PositionHelper.java
https://github.com/checkstyle/checkstyle-openrewrite-recipes/blob/5815f1de75270f974a355e0a7d1e9cdc5b76dd49/src/main/java/org/checkstyle/autofix/PositionHelper.java#L83-L99

* src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
https://github.com/checkstyle/checkstyle-openrewrite-recipes/blob/38afaf33ea984df2253f96c281b7a2cc68616965/src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java#L335-L350

As part of this work:

* Refactor both files to avoid org.openrewrite.internal.RecipeRunException
* Remove temporary ImportControlCheck suppressions after refactoring
* Ensure no internal OpenRewrite APIs are used in the project

https://github.com/checkstyle/checkstyle-openrewrite-recipes/blob/b44d881454e881852fec7a0abbbce65dad97619f/config/suppressions.xml#L34

romani:
issue stays open until we clarify https://github.com/openrewrite/rewrite/issues/7775#issuecomment-4553741906

romani:
@MonuChaudhary14 , please explain usage of this exception. What will happen if we remove such catches ?

MonuChaudhary14:
In both `UseEnhancedSwitch` and `PositionHelper`, we need to traverse the LST but stop immediately when a specific condition is met:
- In `UseEnhancedSwitch`, the `ControlFlowVisitor` stops scanning as soon as it finds restricted control flow (`return`, `break`, `continue`).
- In `PositionHelper`, the overridden `PrintOutputCapture` stops printing text as soon as it reaches the target LST node.

To instantly short-circuit a deeply recursive LST traversal without passing boolean flags all the way up the call stack, we use **Exceptions for Control Flow**. We intentionally throw custom exceptions (`ControlFlowException` and `CancellationException`, respectively) to halt the execution immediately.

OpenRewrite actively intercepts any unchecked exceptions thrown during the `JavaVisitor` or `TreeVisitor` lifecycle. 

[...]

PR #201, opened 2026-05-25
Issue #198: Refractored the usage of internal API

MonuChaudhary14:
Issue #198

### DIFF
--- a/config/import-control.xml
+++ b/config/import-control.xml
@@ -4,6 +4,9 @@
         "http://checkstyle.sourceforge.net/dtds/import_control_1_2.dtd">
 
 <import-control pkg="org.checkstyle.autofix" regex="true">
+    <!-- Temporary allowance until https://github.com/openrewrite/rewrite/issues/7775 -->
+    <allow class="org.openrewrite.internal.RecipeRunException"/>
+
     <disallow pkg="org.openrewrite.internal"/>
     <allow pkg="org.openrewrite"/>
     <allow pkg="org.openrewrite.java"/>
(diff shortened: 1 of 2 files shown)


MonuChaudhary14:
@romani @rdiachenko @sabaka 

I tried multiple approaches to remove this, but the build kept failing.

This code behaves similarly to the previous RecipeRunException case, but here we have an illegal catch, which is not allowed by the Checkstyle configuration

For now, I handled it using this approach and suppressed the illegal catch warning

Please let me know if we can proceed with this approach, or if I should explore another solution

romani:
I think this exception is api problem of openrewrite.
https://github.com/checkstyle/checkstyle-openrewrite-recipes/blob/459c18ae9145592715d4d214b8b3659a5511c163/src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java#L346

it comes from them and we have to use it.

romani:
@MonuChaudhary14 , please open issue against openrewrite to ask them what is prefererable approach.
allow this special import in config with comment above "until XXXX"

MonuChaudhary14:
@romani @sabaka @rdiachenko 

Please check the issue description
https://github.com/openrewrite/rewrite/issues/7775

MonuChaudhary14:
@romani 

Please check this approach?
I have only allowed RecipeRunException

config/import-control.xml
```diff
@@ -4,6 +4,9 @@
         "http://checkstyle.sourceforge.net/dtds/import_control_1_2.dtd">
 
 <import-control pkg="org.checkstyle.autofix" regex="true">
+    <!-- Temporary allowance until https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/198-->
```
romani: Put link to their issue

romani (changes requested):
Items

config/import-control.xml
MonuChaudhary14: done

romani (approved):
lets merge this , as rereview why we need to catch such exception

Issue #202, opened 2026-05-25
Remove FileLength suppressions from test resource input files


MonuChaudhary14:
# Description

Part of parent issue: https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/195

## Removed suppressions

- [x] `InputLocalVariableAssignedMultipleTimes.java`
- [x] `InputLocalVariableCheckSwitchExpressions.java`
- [x] `InputMutationTests.java`
- [x] `InputMutationTests14.java`
- [x] `InputMutationTests15.java`
- [x] `InputMutationTests4.java`
- [x] `InputMutationTests6.java`
- [x] `InputSurvivingMutations.java`
- [x] `InputSurvivingMutations2.java`

## Requirements

- Remove corresponding `FileLength` suppressions from `suppressions.xml`
- Refactor or split files to satisfy `FileLength`
- Ensure all tests pass

romani:
Suppression is clear 
https://github.com/checkstyle/checkstyle-openrewrite-recipes/blob/main/config/checkstyle-resources-suppressions.xml

@MonuChaudhary14 , thanks a lot for clean up

PR #204, opened 2026-05-26
Issue #202: Reduced the file length of LocalVariableAssignedMultipleTimes

MonuChaudhary14:
Issue #202 

I have updated LocalVariableAssignedMultipleTimes.

In this file, the foo11 method has a length of around 180 lines, so I had to split it into multiple parts to satisfy the file length requirement.

I also preserved the comments that were already present in the original file.


romani (approved):
Good step !!!

PR #215, opened 2026-05-29
Issue #198: Refractor the use of internal api

MonuChaudhary14:
Issue #198

### DIFF
--- a/src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
+++ b/src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
@@ -30,7 +30,6 @@ import org.openrewrite.ExecutionContext;
 import org.openrewrite.Recipe;
 import org.openrewrite.Tree;
 import org.openrewrite.TreeVisitor;
-import org.openrewrite.internal.RecipeRunException;
 import org.openrewrite.java.JavaVisitor;
 import org.openrewrite.java.tree.Expression;
 import org.openrewrite.java.tree.Flag;
@@ -333,20 +332,11 @@ public class UseEnhancedSwitch extends Recipe {
         }
 
         private boolean hasDisallowedControlFlow(final List<Statement> stmts) {
-            boolean hasFlow = false;
             final ControlFlowVisitor visitor = new ControlFlowVisitor();
-            try {
-                for (int index = 0; index < stmts.size() - 1; index++) {
-                    visitor.visit(stmts.get(index), null);
-                }
+            for (int index = 0; index < stmts.size() - 1; index++) {
+                visitor.visit(stmts.get(index), null);
             }
-            catch (ControlFlowException exception) {
-                hasFlow = true;
-            }
-            catch (RecipeRunException exception) {
-                hasFlow = true;
-            }
-            return hasFlow;
+            return visitor.hasDisallowedFlow();
         }
 
         private List<Statement> getMeaningfulStatements(final J.Case caseStmt) {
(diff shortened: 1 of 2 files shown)


MonuChaudhary14:
@romani 

Please check this, then and tell me if this can work or not?

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -653,24 +646,32 @@ private static int getIndentLevel(final Space space) {
...
 
-        private static final class ControlFlowException extends RuntimeException {
-            private static final long serialVersionUID = 1L;
-        }
-
         private static final class ControlFlowVisitor extends JavaVisitor<Void> {
+            private boolean hasFlow;
+
+            public boolean hasDisallowedFlow() {
+                return hasFlow;
```
romani: make field to match getter name. 

`private boolean hasDisallowedFlow;`

romani (changes requested):
I think it is better to split this PR into two and fix both usecases separately, it will speedup review.

items for first:

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
MonuChaudhary14: @romani 

I am stuck in one place

There are some mutations of UseEnhancedSwitch, for which to kill i have to return null from the method, but the OpenRewrite best practices are adding `@Nullable` in it, but using Nullable is not allowed in check, 

What should i do, because other approach will require to throw exceptions, which we are avoiding and for Nullable, i have to suppress the ImportControlCheck, what should i do

romani: In your code , let's return not a null , but object from parameters. As you do no action.

In pitest we can exclude from mutation inner classes by name. Loook at main repository for examples.

pom.xml
```diff
@@ -405,6 +405,9 @@
                             </dependency>
                         </dependencies>
                         <configuration>
+                            <excludedClasses>
+                                <param>org.checkstyle.autofix.recipe.UseEnhancedSwitch$UseEnhancedSwitchVisitor$ControlFlowVisitor</param>
```
romani: Ok, this workaround.
Please share list of suppressions that are required without it.

romani (changes requested):
Items

pom.xml
MonuChaudhary14:
```
  <mutation unstable="false">
    <sourceFile>UseEnhancedSwitch.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.recipe.UseEnhancedSwitch$UseEnhancedSwitchVisitor$ControlFlowVisitor</mutatedClass>
    <mutatedMethod>visitBreak</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.NullReturnValsMutator</mutator>
    <description>replaced return value with null for org/checkstyle/autofix/recipe/UseEnhancedSwitch$UseEnhancedSwitchVisitor$ControlFlowVisitor::visitBreak</description>
    <lineContent>return breakObj;</lineContent>
  </mutation>
```
```
  <mutation unstable="false">
    <sourceFile>UseEnhancedSwitch.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.recipe.UseEnhancedSwitch$UseEnhancedSwitchVisitor$ControlFlowVisitor</mutatedClass>
    <mutatedMethod>visitContinue</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.NullReturnValsMutator</mutator>
    <description>replaced return value with null for org/checkstyle/autofix/recipe/UseEnhancedSwitch$UseEnhancedSwitchVisitor$ControlFlowVisitor::visitContinue</description>
    <lineContent>return continueObj;</lineContent>
  </mutation>
```
```
  <mutation unstable="false">
    <sourceFile>UseEnhancedSwitch.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.recipe.UseEnhancedSwitch$UseEnhancedSwitchVisitor$ControlFlowVisitor</mutatedClass>
    <mutatedMethod>visitReturn</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.NullReturnValsMutator</mutator>
    <description>replaced return value with null for org/checkstyle/autofix/recipe/UseEnhancedSwitch$UseEnhancedSwitchVisitor$ControlFlowVisitor::visitReturn</description>
    <lineContent>return returnObj;</lineContent>
  </mutation>
```

romani: @MonuChaudhary14 ,
please remove this exclude , and let accept new pitest mutaitons.

Looks like we will reconsider all pitest after migration to markers.

MonuChaudhary14: done

MonuChaudhary14: @romani 

Ping

romani (approved):
Please recheck that all suppressions are removed

Issue #234, opened 2026-06-09
Migrate all the remaining recipes to marker based approach


MonuChaudhary14:
# Description

We have to migrate all the recipes to marker based approach

- [x] AnnotationOnSameLine
- [x] AvoidStarImport
- [x] ConstructorDeclarationGrouping
- [x] Empty Statement
- [x] Header
- [x] HexLiteralCase
- [x] MissingDeprecated
- [x] MissingOverride
- [x] NewLineAtEndOfFile
- [x] NumericalPrefixesInfixesSuffixesCharacterCase
- [x] RedundantImport
- [x] UnusedImports
- [x] UnusedLocalVariable
- [x] UpperEll
- [x] UseEnhancedSwitch

rdiachenko:
all existing recipes migrated, thanks a lot!

PR #239, opened 2026-06-11
Issue #234: Migrate the ConstructorsDeclarationGrouping recipe into marker

### LINKED ISSUE #175: Bug: ConstructorsDeclarationGrouping wrongly moves suppressed constructors at second cycle
Problem
---
With the current position-based violation locating mechanism, ConstructorsDeclarationGrouping sometimes wrongly moves ungrouped constructors that are suppressed. 

Example
---
**input code:**
```java
public class Test {
    Test() {}
    Test(String s) {}
    void foo() {}
    Test(int x) {}
    Test(String s, int x) {} 
}
```
**config:**
```xml
<?xml version="1.0"?>
<!DOCTYPE module PUBLIC
        "-//Checkstyle//DTD Checkstyle Configuration 1.3//EN"
        "https://checkstyle.org/dtds/configuration_1_3.dtd">
<module name="Checker">
    <module name="TreeWalker">
        <module name="ConstructorsDeclarationGrouping"/>
        <module name="SuppressionXpathSingleFilter">
            <property name="checks"
                      value="ConstructorsDeclarationGrouping"/>
            <property name="query"
                      value="//CTOR_DEF[./PARAMETERS/PARAMETER_DEF/TYPE/LITERAL_INT and not(./PARAMETERS/PARAMETER_DEF/TYPE/IDENT)]"/>
        </module>
    </module>

</module>
```
Note: the filter suppresses the constructor `Test(int x)`.

**checkstyle violation report:**
```
Starting audit...
[ERROR] /mnt/d/IntellijProjects/checkstyle-playground/inputs/constructors-declaration-grouping/Test.java:6:5: Constructors should be grouped together. The last grouped constructor is declared at line '3'. [ConstructorsDeclarationGrouping]
Audit done.
Checkstyle ends with 1 errors.
```

Output after first recipe run
---
```java
public class Test{
    Test() {}
    Test(String s) {}
    Test(String s, int x) {}
    void foo() {}
    Test(int x) {}
}
```
This is the desired output. Unfortunately, `Test(int x)` is now at the violation location (6:5: ), thus will be wrongly moved after the second cycle.

Output after second recipe run
---
```java
public class Test{
    Test() {}
    Test(String s) {}
    Test(String s, int x) {}
    Test(int x) {}
    void foo() {}
}
```

from issue: #167

lithops-zty:
Issue: #234
fixes #175 
The two error inputs now pass and are unsuppressed.


src/test/java/org/checkstyle/autofix/marker/ViolationMarkerRecipeTest.java
```diff
@@ -898,6 +898,43 @@ public void testResolveTargetNodeFinalClassSwitch() throws Exception {
...
+        nodeTrees.put(childId, cu);
+        nodeTrees.put(constructorId, constructor);
+        final CheckstyleCheck checkCdg = new CheckstyleCheck(
+                CheckFullName.CONSTRUCTORS_DECLARATION_GROUPING, "id");
+        final CheckstyleViolation violCdg = new CheckstyleViolation(
+                1, 1, "err", checkCdg, "msg", Paths.get("a"));
+        final UUID resultCdg = (UUID) resolveMethod.invoke(
+                scanner, childId, violCdg, parentMap, nodeTrees);
+        Assertions.assertEquals(constructorId, resultCdg,
+                "CONSTRUCTORS_DECLARATION_GROUPING should resolve up to MethodDeclaration");
```
romani: Why we can not cover this by input files?

romani (changes requested):
We should do tests by Input files as we do in main repository

src/test/java/org/checkstyle/autofix/marker/ViolationMarkerRecipeTest.java
lithops-zty: @romani The unit test was a workaround to kill mutation on `ViolationMarkerRecipe`. This class is in pitest-coding-1 but the `ConstructorDeclarationGrouping` recipe is in pitest-coding-2. Recipe tests are required to kill mutations on recipe-specific code, but are never run due to the profile split.
PR #241 solves this by adding a pitest-internal profile that tests ViolationMarkerRecipe against all tests. The reflection test can be removed after #241 is merged.

MonuChaudhary14:
@lithops-zty 

Please rebase your PR with latest changes
Also, remove the unit test, if required, add the test in pitest-marker profile as you mentioned above

config/pitest-recipe-2-suppressions.xml
```diff
@@ -20,6 +20,33 @@
...
+  </mutation>
+
+  <mutation unstable="false">
+    <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
+    <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
+    <mutatedMethod>findViolatingConstructors</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
+    <description>removed conditional - replaced equality check with true</description>
+    <lineContent>&amp;&amp; method.isConstructor()</lineContent>
+  </mutation>
```
MonuChaudhary14, line 39: Please remove these two suppressions

config/pitest-recipe-2-suppressions.xml
```diff
@@ -20,6 +20,33 @@
...
+
```
MonuChaudhary14, line 40: Add this comment to before the 3rd suppressed mutation identify better, why it is suppressed
```<!-- Cannot be killed by integration tests until a second recipe lands a marker on this node type. See issue #236 -->```

**Comment by romani** [comment]
@lithops-zty , please do suggested updates.

almost ready to be merged

**Comment by lithops-zty** [comment]
done

**Review by romani — APPROVED** [review]
@MonuChaudhary14 , please finalize review

File: src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java [hunk 4]
```diff
@@ -211,9 +211,9 @@ private UUID resolveTargetNode(UUID startNodeId, CheckstyleViolation violation,
                     case FINAL_CLASS -> J.ClassDeclaration.class;
                     case FINAL_LOCAL_VARIABLE -> J.VariableDeclarations.NamedVariable.class;
                     case ANNOTATION_ON_SAME_LINE -> J.Annotation.class;
-                    case AVOID_STAR_IMPORT, UNUSED_IMPORT -> J.Import.class;
-                    case UNUSED_LOCAL_VARIABLE -> J.VariableDeclarations.class;
+                    case AVOID_STAR_IMPORT -> J.Import.class;
                     case MISSING_OVERRIDE -> J.MethodDeclaration.class;
+                    case CONSTRUCTORS_DECLARATION_GROUPING -> J.MethodDeclaration.class;
```
MonuChaudhary14: Please merge the `MISSING_OVERRIDE` and `CONSTRUCTORS_DECLARATION_GROUPING` into a single case, as they are returning the same value

`case MISSING_OVERRIDE, CONSTRUCTORS_DECLARATION_GROUPING -> J.MethodDeclaration.class;`

@romani 
There is no check in checkstyle which detects the duplicate branch in switch, and also no Open-rewrite recipe for it, 
Can we make a new Check for this

src/main/java/org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping.java
```diff
@@ -111,18 +95,44 @@ private static int findFirstConstructorIndex(List<Statement> statements) {
...
+        private static int findFirstConstructorGroupEnd(List<Statement> statements) {
+            int result = findFirstConstructorIndex(statements);
+            if (result != -1) {
+                while (statements.get(result) instanceof J.MethodDeclaration method
+                        && method.isConstructor()) {
+                    result++;
+                }
+            }
+            return result;
+        }
```
MonuChaudhary14, line 126: Can you add a test in which the last statement is constructor, 
I think there is a bug in while loop which can cause `IndexOutOfBoundException`

```
while (statements.get(result) instanceof J.MethodDeclaration method
    && method.isConstructor()) {
    result++;
}
```
If a class's very last statement happens to be a constructor (a class that contains only constructors), result will increment until it equals statements.size(). The while loop will then attempt to evaluate statements.get(result), which will immediately throw an IndexOutOfBoundsException because it's trying to access an index outside the bounds of the array.

Also, check the `reorderConstructors` method


Try with this Input and Ouput, and add them
Input:
```
public class InputClassEndsWithConstructor {

    InputClassEndsWithConstructor(int x) {}

    void foo() {}

    InputClassEndsWithConstructor() {} // violation 'Constructors should be grouped together.*'

    static class AnotherClass {

        AnotherClass() {}

        AnotherClass(int x) {} // ends with constructor
    }
}
```

Output:
```
public class OutputClassEndsWithConstructor {

    OutputClassEndsWithConstructor(int x) {}

    OutputClassEndsWithConstructor() {}

    void foo() {}

    static class AnotherClass {

        AnotherClass() {}

        AnotherClass(int x) {}
    }
}
```

MonuChaudhary14 (changes requested):
Please, make these changes

src/main/java/org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping.java
lithops-zty: Thank you I wasn't able to find this case where pointer increments beyond last.

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
romani: Sure, please create issue

lithops-zty: done

src/main/java/org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping.java
MonuChaudhary14: Can you make a small change in this 
```
if (firstConstructorIndex == -1 || violatingConstructors.isEmpty()) {
    newClassDeclaration = classDeclaration;
}
```

Currently, if a class has no violations, the code still creates a new ArrayList, assign it and calls the classDeclaration.withBody method

In open-rewrite , returning a newly allocated list makes the framework think the AST was modified
This can cause OpenRewrite to run unnecessary processing cycles

MonuChaudhary14 (changes requested):
Please made this minor change

src/main/java/org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping.java
MonuChaudhary14: @romani 

Can you resolve this comment as i don't have right access
the bug is fixed

src/main/java/org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping.java
lithops-zty: No problem! This is a good performance booster I did not think of.
However I suspect PITest will complain saying removing this condition does not lead to wrong output. In this case can I add a suppression?

MonuChaudhary14: Can you share which mutation is surviving? That would help determine whether a suppression is actually needed or if there’s a way to refactor the code so the mutation is killed naturally

MonuChaudhary14:
```
  <mutation unstable="false">
    <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
    <mutatedMethod>visitClassDeclaration</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_ELSE</mutator>
    <description>removed conditional - replaced equality check with false</description>
    <lineContent>if (firstConstructorIndex == -1 || violatingConstructors.isEmpty()) {</lineContent>
  </mutation>

  <mutation unstable="false">
    <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
    <mutatedMethod>visitClassDeclaration</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
    <description>removed conditional - replaced equality check with true</description>
    <lineContent>if (firstConstructorIndex == -1 || violatingConstructors.isEmpty()) {</lineContent>
  </mutation>

<mutation unstable="false">
    <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
    <mutatedMethod>visitClassDeclaration</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
    <description>removed call to java/util/List::isEmpty</description>
    <lineContent>if (firstConstructorIndex == -1 || violatingConstructors.isEmpty()) {</lineContent>
  </mutation>

  <mutation unstable="false">
    <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
    <mutatedMethod>reorderConstructors</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_ORDER_IF</mutator>
    <description>removed conditional - replaced comparison check with true</description>
    <lineContent>while (insertIndex &lt; statements.size()</lineContent>
  </mutation>
```

@romani 
What do you think about this, should be suppress them or we can leave this `|| violatingConstructors.isEmpty()` part with some comments to improve in future

lithops-zty:
```xml
<mutation unstable="false">
  <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
  <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
  <mutatedMethod>visitClassDeclaration</mutatedMethod>
  <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
  <description>removed conditional - replaced equality check with true</description>
  <lineContent>if (firstConstructorIndex == -1 || violatingConstructors.isEmpty()) {</lineContent>
</mutation>
```
I have no idea why this survived. Doesn't it always skip reordering, making the recipe useless? Am I missing something

romani: we are ok to use suppression for pitest to finish migration to marker.

We will give a fight for all pitest after: marker is done and diff report is ready to use

config/pitest-recipe-2-suppressions.xml
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <!-- Cannot be killed by integration tests until a second recipe lands a marker on this node type. See issue #236 -->
```
MonuChaudhary14, line 3: Please add these comments back, and this PR will be ready to merge.

MonuChaudhary14 (approved):

MonuChaudhary14:
@romani 

Please merge this PR

romani:
@lithops-zty , please resolve conflict

Issue #231, opened 2026-06-08
Move violation target resolution into CheckFullName enum (predicate per check)


MonuChaudhary14:
# Description
Currently, in `ViolationMarkerRecipe`, we use a `switch` statement inside the `ScannerVisitor#resolveTargetNode` method to walk up the AST and find the correct target container for a violation marker based on the `CheckFullName`. 
For example, `FINAL_CLASS` walks up to find a `J.ClassDeclaration` and `FINAL_LOCAL_VARIABLE` walks up to find a `J.VariableDeclarations.NamedVariable`.
As we migrate more recipes, this `switch` statement will grow large and difficult to maintain. To improve the design and encapsulate the logic, we should move this target resolution logic directly into the `CheckFullName` enum itself.

rdiachenko:
@MonuChaudhary14 we may not need to do this refactoring anymore, but let's do clean up in scope of this issue. please check and remove old logic, which is now covered by markers (we have position helper class, there may be other leftovers)

rdiachenko:
fix is merged

PR #242, opened 2026-06-13
Issue #234: Migrated the empty Statement recipe to marker

MonuChaudhary14:
Issue #234

### DIFF
--- a/config/pitest-coding-1-suppressions.xml
+++ b/config/pitest-coding-1-suppressions.xml
@@ -173,21 +173,22 @@
 
   <!-- Cannot be killed by integration tests until a second recipe lands a marker on this node type. See issue #236 -->
   <mutation unstable="false">
+    <sourceFile>EmptyStatement.java</sourceFile>
+    <mutatedClass>org.checkstyle.autofix.recipe.EmptyStatement$EmptyStatementVisitor</mutatedClass>
+    <mutatedMethod>lambda$isAtViolationLocation$1</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.BooleanTrueReturnValsMutator</mutator>
+    <description>replaced boolean return with true for org/checkstyle/autofix/recipe/EmptyStatement$EmptyStatementVisitor::lambda$isAtViolationLocation$1</description>
+    <lineContent>.anyMatch(marker -&gt; marker.isFor(CheckFullName.EMPTY_STATEMENT));</lineContent>
+  </mutation>
+
+  <!-- Cannot be killed by integration tests until a second recipe lands a marker on this node type. See issue #236 -->
+  <mutation unstable="false">
     <sourceFile>FinalClass.java</sourceFile>
     <mutatedClass>org.checkstyle.autofix.recipe.FinalClass$FinalClassVisitor</mutatedClass>
     <mutatedMethod>lambda$visitClassDeclaration$0</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.BooleanTrueReturnValsMutator</mutator>
     <description>replaced boolean return with true for org/checkstyle/autofix/recipe/FinalClass$FinalClassVisitor::lambda$visitClassDeclaration$0</description>
     <lineContent>.anyMatch(marker -&gt; marker.isFor(CheckFullName.FINAL_CLASS));</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>FinalLocalVariable.java</sourceFile>
-    <mutatedClass>org.checkstyle.autofix.recipe.FinalLocalVariable$LocalVariableVisitor</mutatedClass>
-    <mutatedMethod>lambda$isViolationMarked$0</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.BooleanTrueReturnValsMutator</mutator>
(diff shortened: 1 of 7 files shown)


MonuChaudhary14:
@romani @rdiachenko 

After switching to marker based approach, we are getting a PIT mutation which is 
```
<mutation unstable="false">
  <sourceFile>ViolationMarkerRecipe.java</sourceFile>
  <mutatedClass>org.checkstyle.autofix.marker.ViolationMarkerRecipe$ScannerVisitor</mutatedClass>
  <mutatedMethod>resolveTargetNode</mutatedMethod>
  <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.RemoveSwitchMutator_2</mutator>
  <description>RemoveSwitch 2 (case value 3)</description>
  <lineContent>final Class&lt;? extends Tree&gt; targetType = switch (checkFullName) {</lineContent>
</mutation>
```
The PIT is removing the 3rd case
```
final Class<? extends Tree> targetType = switch (checkFullName) {
    case FINAL_CLASS -> J.ClassDeclaration.class;
    case FINAL_LOCAL_VARIABLE -> J.VariableDeclarations.NamedVariable.class;
    case EMPTY_STATEMENT -> J.Empty.class;
    default -> null;
};
```

The PIT mutator was removing the `case EMPTY_STATEMENT -> J.Empty.class;` branch from the switch statement.
When an `EMPTY_STATEMENT` violation occurs, it points to the exact line/col of a standalone semicolon. The smallest node at that location is already `J.Empty`. When the PIT mutator removed the `EMPTY_STATEMENT` case from the switch, the code fell through to the `default -> null` branch, which just returns the original startNodeId (the smallest node). Because startNodeId was already `J.Empty,` the mutated code and the original code produced the exact same output. Since the behavior was identical, the tests continued to pass, and the mutation survived.

Temporary Solution:
Because the `EMPTY_STATEMENT` case was functionally redundant, I simply removed it from the switch statement entirely. The default fallback works perfectly for it, all EmptyStatementTests still pass, and the mutation is now fully eliminated. Going forward, we will need to make sure we add specific unit tests for most of the recipes we add to this switch block to ensure their switch cases are properly covered and don't allow PIT mutations to survive.

Regarding your previous direction:
`"Move violation target resolution into CheckFullName enum (predicate per check)". We will move the switch logic into the enum once a few more recipes are migrated.`
https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/231

We can switch to this approach now to remove this class of mutations for all future recipes. 
By adding a `targetNodeType` field and `getTargetNodeType()` method directly to the CheckFullName enum, we can delete the switch statement entirely and permanently prevent RemoveSwitch mutations here.

rdiachenko:
@MonuChaudhary14 let's go with temporary solution, if swtich case produces the same node as default one, let's not have the case.

Regarding 
> "Move violation target resolution into CheckFullName enum (predicate per check)". We will move the switch logic into the enum once a few more recipes are migrated.

Let's do this after all current recipes are migrated. We need to think whether CheckFullName is the best place for such a logic. We need more data from migration to decide.

MonuChaudhary14:
@rdiachenko 
Please check this

romani:
I am fine to temporarily extend pitest mutations

config/pitest-coding-1-suppressions.xml
```diff
@@ -532,6 +533,15 @@
     <lineContent>column++;</lineContent>
   </mutation>
 
+  <mutation unstable="false">
+    <sourceFile>PositionHelper.java</sourceFile>
```
rdiachenko, line 547: It's fine, we should remove this class after migration to marker-based approach

src/main/java/org/checkstyle/autofix/recipe/EmptyStatement.java
```diff
@@ -150,15 +135,12 @@ && isAtViolationLocation((J.Empty) statement)) {
...
-                return violation.getFilePath().endsWith(compilationUnit.getSourcePath())
-                        && violation.getLine() == line
-                        && violation.getColumn() == column;
-            });
+            boolean found = false;
+            for (CheckstyleViolationMarker marker : empty.getMarkers()
+                    .findAll(CheckstyleViolationMarker.class)) {
+                found |= marker.isFor(CheckFullName.EMPTY_STATEMENT);
+            }
+            return found;
```
rdiachenko:
```
return empty.getMarkers()
            .findAll(CheckstyleViolationMarker.class)
            .stream()
            .anyMatch(marker -> marker.isFor(CheckFullName.EMPTY_STATEMENT));
```

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -140,17 +140,30 @@ private Map<UUID, List<CheckstyleViolationMarker>> findSmallestNode(
...
                         minCols = cols;
                         smallestNodeId = entry.getKey();
                     }
+                    else if (exactMatch || !foundExactMatch) {
+                        if (lines < minLines || lines == minLines && cols <= minCols) {
+                            minLines = lines;
+                            minCols = cols;
+                            smallestNodeId = entry.getKey();
+                        }
+                    }
```
rdiachenko: can we simplify the logic here, since most is duplicated?

src/main/java/org/checkstyle/autofix/recipe/EmptyStatement.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

rdiachenko (approved):
lgtm

PR #244, opened 2026-06-15
Issue #234: Migrated the Header recipe into marker

MonuChaudhary14:
Issue #234

### DIFF
--- a/config/pitest-coding-1-suppressions.xml
+++ b/config/pitest-coding-1-suppressions.xml
@@ -175,7 +175,43 @@
   <mutation unstable="false">
     <sourceFile>CheckstyleRecipeRegistry.java</sourceFile>
     <mutatedClass>org.checkstyle.autofix.CheckstyleRecipeRegistry</mutatedClass>
+    <mutatedMethod>addRecipe</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
+    <description>removed call to java/util/EnumMap::get</description>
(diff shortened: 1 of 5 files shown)


src/test/java/org/checkstyle/autofix/marker/ViolationMarkerRecipeTest.java
```diff
@@ -217,6 +217,57 @@ else if (tree instanceof J.Block) {
         Assertions.assertFalse(blockHasMarker[0], "Block should NOT have the marker");
     }
 
+    @Test
+    public void testMultiLineNodeEndsBeforeViolationOnSameLine() {
```
rdiachenko: what's the reason for these unit tests? you updated the logic in the marker recipe specifically for the Header recipe, so why can't we cover it by new inputs?

MonuChaudhary14: I dont know, how I miss this, yes the tests are already covering these mutations, i have removed the unit tests

MonuChaudhary14: I think, now we can remove some unit tests from ViolationMarkerRecipeTest, as so many recipes has been migrated to marker

rdiachenko (approved):
this is the first recipe that goes into its own configuration map, so new mutations in `CheckstyleRecipeRegistry`. they should be fixed when more recipes will be migrated to use marker and similar config map. and all leftovers will be fixed as part of https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/183

ok to merge

PR #249, opened 2026-06-17
Issue  #234: Migrated the MissingDeprecated Recipe to marker

MonuChaudhary14:
Issue #234

### DIFF
--- a/src/main/java/org/checkstyle/autofix/recipe/MissingDeprecated.java
+++ b/src/main/java/org/checkstyle/autofix/recipe/MissingDeprecated.java
@@ -17,13 +17,12 @@
 
 package org.checkstyle.autofix.recipe;
 
-import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.stream.IntStream;
 
-import org.checkstyle.autofix.PositionHelper;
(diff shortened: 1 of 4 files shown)


src/main/java/org/checkstyle/autofix/recipe/MissingDeprecated.java
```diff
@@ -67,27 +60,22 @@ public TreeVisitor<?, ExecutionContext> getVisitor() {
...
-
         @Override
         public J.MethodDeclaration visitMethodDeclaration(J.MethodDeclaration method,
                 ExecutionContext executionContext) {
             J.MethodDeclaration result = method;
 
-            if (isAtViolationLocation(result)) {
+            final boolean hasMarker = result.getMarkers()
+                    .findAll(CheckstyleViolationMarker.class).stream()
+                    .anyMatch(marker -> marker.isFor(CheckFullName.MISSING_DEPRECATED));
```
rdiachenko: since this is duplicated accross multiple places in this class, let's move it into its own method and reuse

MonuChaudhary14: done

rdiachenko (approved):
lgtm

Issue #237, opened 2026-06-11
Remove the unit tests from ViolationMarkerRecipeTest


MonuChaudhary14:
# Description
Remove the unit test from ViolationMarkerRecipeTest, without creating more mutations

PR #253, opened 2026-06-19
Issue #234: Fixed the bug caused by migration of recipe to marker

MonuChaudhary14:
Issue #234


MonuChaudhary14:
@rdiachenko 

Please check this

rdiachenko (commented):
1. update description with a link to issue
2. why unit test? we should always aim to cover logic by having proper inputs

MonuChaudhary14:
> 1. update description with a link to issue
> 2. why unit test? we should always aim to cover logic by having proper input

The reason I have added the unit test is that the changes I made in ViolationMarkerRecipe require `violation.getColumn() <= 0` to be true. However, there is no recipe tested in the pitest-coding-1 profile that generates a negative or missing column number, so the condition is always false during that PIT run

The recipe which provides proper input to kill this mutation is MissingDeprecated, because it doe not have any column in the violation, but that test is present in the pitest-coding-2 profile.

rdiachenko:
Ok, let's merge it to unblock main branch. We have issue where we should remove all UTs for ViolationMarkerRecipe https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/237 after migration is done

rdiachenko (approved):

PR #267, opened 2026-06-22
Issue #234: Migrated the UpperEll recipe to marker

### LINKED ISSUE #234: Migrate all the remaining recipes to marker based approach

MonuChaudhary14:
Fixes #234 

this is final fix for migration to Marker model.

### DIFF
--- a/src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
+++ b/src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
@@ -20,6 +20,7 @@ package org.checkstyle.autofix.marker;
 import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.Iterator;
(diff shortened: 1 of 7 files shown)


MonuChaudhary14:
@sabaka 

Can you review this PR, as we are very close to migrate all the recipes to marker

rdiachenko:
@MonuChaudhary14 please resolve conflicts

MonuChaudhary14:
> @MonuChaudhary14 please resolve conflicts

done

romani:
@MonuChaudhary14 , one more conflict.

MonuChaudhary14:
okay, i fix this

MonuChaudhary14:
> @MonuChaudhary14 , one more conflict.

done

MonuChaudhary14:
@sabaka @rdiachenko 

Can you review this last PR of migration

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -173,45 +174,44 @@ private Map<UUID, List<CheckstyleViolationMarker>> findSmallestNode(
...
+                    .reduce((prev, next) -> {
+                        final Map.Entry<UUID, Range> better;
+                        if (cmp.compare(prev, next) < 0) {
+                            better = prev;
+                        }
+                        else {
+                            better = next;
+                        }
+                        return better;
+                    })
```
rdiachenko, line 210: could you check if we can simplify this reduce and replace it with just `.min(cmp)`

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -173,45 +174,44 @@ private Map<UUID, List<CheckstyleViolationMarker>> findSmallestNode(
...
+                    .filter(entry -> {
+                        final boolean matches;
+                        if (lineOnly) {
+                            matches = enclosesLine(entry.getValue(), violation);
+                        }
+                        else {
+                            matches = encloses(entry.getValue(), violation);
+                        }
+                        return matches;
+                    })
```
rdiachenko, line 200: can we simplify this to:
```
.filter(entry -> lineOnly
          ? enclosesLine(entry.getValue(), violation)
          : encloses(entry.getValue(), violation))
```

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -296,6 +297,9 @@ private static final class BoundingBoxCapture extends PrintOutputCapture<TreeVis
 
         @Override
         public PrintOutputCapture<TreeVisitor<?, ?>> append(String text) {
+            final String skipped = out.substring(lastLength);
+            column += skipped.length();
```
rdiachenko: `skipped` may contain a new line, so the calculation for column may be incorrect in this case. we need to apply similar logic as you have below. let's move it into a helper method and reuse, something like:
```
  private void advance(String segment) {
      for (int index = 0; index < segment.length(); index++) {
          if (segment.charAt(index) == '\n') {
              line++;
              column = 1;
          }
          else {
              column++;
          }
      }
  }

  @Override
  public PrintOutputCapture<TreeVisitor<?, ?>> append(String text) {
      advance(out.substring(lastLength));   // skipped, now newline-aware
      ...
      advance(text);                        // replaces the inline loop
      ...
      super.append(text);
      lastLength = out.length();
      return this;
  }
```

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -282,6 +282,7 @@ private static final class BoundingBoxCapture extends PrintOutputCapture<TreeVis
         private final Map<UUID, Tree> nodeTrees;
         private int line;
         private int column;
+        private int lastLength;
```
rdiachenko: coulde rename it to `lastAppendEnd`, and let's leave a comment for why we need it, since it is not obvious:
```
  // Index in out where last append(String) ended. The printer may also write
  // to out directly (bypassing this override), so anything past this offset is
  // output we haven't yet counted into line/column.
  private int lastAppendEnd;
```

rdiachenko (changes requested):

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
MonuChaudhary14: The checkstyle blocks it, we cannot use InlineConditions (AvoidInlineConditionals)

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
MonuChaudhary14: We also cannot use .min() because when two node ranges are exactly equal, min() returns the first element it encountered (the parent node). 
.reduce() logic returns the second element (the child node), which is required to find the most deeply nested node

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

rdiachenko (approved):
lgtm

PR #268, opened 2026-06-23
Issue #234: Migrated the UseEnhancedRecipe to marker

MonuChaudhary14:
Issue #234

### DIFF
--- a/src/test/resources/org/checkstyle/autofix/recipe/useenhancedswitch/survivingmutations3/DiffSurvivingMutations3.diff
+++ b/src/test/resources/org/checkstyle/autofix/recipe/useenhancedswitch/survivingmutations3/DiffSurvivingMutations3.diff
@@ -6,7 +6,7 @@
  
 -public class InputSurvivingMutations3 {
 +public class OutputSurvivingMutations3 {
-     
+ 
      void m6(int x) {
 -          // violation below, 'Switch can be replaced with enhanced switch'
            switch (x) {
@@ -39,7 +39,7 @@
 -        // violation below, 'Switch can be replaced with enhanced switch'
 -        switch (x) { case 1: switch (y) { case 1: System.out.println("1"); case 2: break; default: break; } break;
 -            default: break;
-+        switch (x) { case 1 -> { switch (y) { case 1 -> System.out.println("1"); case 2 -> {} default -> {} }}
++        switch (x) { case 1 -> { switch (y) { case 1: System.out.println("1"); case 2: break; default: break; }}
 +            default -> {}
          }
      }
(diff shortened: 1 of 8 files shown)


MonuChaudhary14:
@sabaka 

Can you review this PR, as we are very close to migrate all the recipes to marker

src/test/resources/org/checkstyle/autofix/recipe/useenhancedswitch/survivingmutations3/DiffSurvivingMutations3.diff
```diff
@@ -39,7 +39,7 @@
 -        // violation below, 'Switch can be replaced with enhanced switch'
 -        switch (x) { case 1: switch (y) { case 1: System.out.println("1"); case 2: break; default: break; } break;
 -            default: break;
-+        switch (x) { case 1 -> { switch (y) { case 1 -> System.out.println("1"); case 2 -> {} default -> {} }}
++        switch (x) { case 1 -> { switch (y) { case 1: System.out.println("1"); case 2: break; default: break; }}
 +            default -> {}
```
rdiachenko, line 43: what's the reason for this change? we don't expect regression in test data

MonuChaudhary14: In the previous logic, there was a bug,there is only one violation which is for outer switch , when we switched to marker, it correctly attaches the marker to correct switch

rdiachenko: could you recheck why checkstyle didn't give violation for internal switch? is this expected or it's a bug in checkstyle?

MonuChaudhary14: okay, i will check that

MonuChaudhary14: This is the expected behaviour, as the inner switch case 1 does not have break so it falls through to case 2, and enhance switch do not support fall through behavior

rdiachenko: could you add a test input which will contain two violations: one in the outer switch and one in the internal switch. you may just extend this same input file

MonuChaudhary14: done, please check 
https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/268/changes#diff-2324834e3a3ff26c58a41567f9b1b64669df5b2c5bd965edc4af7b4e1cd5a389R66

rdiachenko (approved):
lgtm

Issue #269, opened 2026-06-23
Resolve JUnit version mismatch and checkstyle Constructor Ordering violation


MonuChaudhary14:
Fix the build failure occured in these PRs by dependabot

https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/255
https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/257
https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/258

romani:
all merged.

PR #270, opened 2026-06-24
Issue #269: Fixed the JUnit version mismatch and checkstyle Constructor Ordering violation

MonuChaudhary14:
Issue #269 

In the PRs, created by dependabot, there is a version mismatch between junit-jupiter and junit-platform-engine.
For this PR, https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/257
This is caused by this dependency:
```
<dependency>
    <groupId>org.openrewrite</groupId>
    <artifactId>rewrite-test</artifactId>
    <scope>test</scope>
</dependency>
```
When maven built the project, it looked at rewrite-test first and noticed that it requires junit-platform-engine version 1.13.3

But as the Junit Jupiter version is upgraded to 6.1.0, it requires higher version, and this caused the issue

To fix this , i have added the Junit-bom which forces to use the latest version by dropping the one by rewrite-test

And for this PR, https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/255
rewrite-test pulled a newer version of Junit platform but pom includes Junit-jupiter version as 5.13.4, which looks for previous version of Junit Platform

For this, PR https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/258
It is failing because of ConstructorGrouping Check
```
Error:  src/main/java/org/checkstyle/autofix/parser/CheckstyleViolation.java:[48,5] (coding) ConstructorsDeclarationGrouping: Constructors should be ordered by increasing parameter count.
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -61,6 +61,13 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>${junit.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 
@@ -118,13 +125,11 @@
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter</artifactId>
-            <version>${junit.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-params</artifactId>
-            <version>${junit.version}</version>
             <scope>test</scope>
         </dependency>
         <dependency>
--- a/src/main/java/org/checkstyle/autofix/parser/CheckstyleViolation.java
+++ b/src/main/java/org/checkstyle/autofix/parser/CheckstyleViolation.java
@@ -35,6 +35,11 @@ public final class CheckstyleViolation {
 
     private final Path filePath;
 
+    public CheckstyleViolation(int line, String severity,
+                               CheckstyleCheck source, String message, Path filePath) {
+        this(line, -1, severity, source, message, filePath);
+    }
+
     public CheckstyleViolation(int line, int column, String severity,
                                CheckstyleCheck source, String message, Path filePath) {
         this.line = line;
@@ -43,11 +48,6 @@ public final class CheckstyleViolation {
         this.source = source;
         this.message = message;
         this.filePath = filePath;
-    }
-
-    public CheckstyleViolation(int line, String severity,
-                               CheckstyleCheck source, String message, Path filePath) {
-        this(line, -1, severity, source, message, filePath);
     }
 
     public Integer getLine() {


MonuChaudhary14:
@romani 

Please check this

rewrite.yml
```diff
@@ -51,7 +51,7 @@ recipeList:
   - org.openrewrite.java.recipes.NoMutableStaticFieldsInRecipes
   - org.openrewrite.java.recipes.RecipeEqualsAndHashCodeCallSuper
   - org.openrewrite.java.recipes.UseTreeRandomId
-  - org.openrewrite.java.InlineMethodCalls
+#  - org.openrewrite.java.InlineMethodCalls
```
rdiachenko: what's the reason? please revert

pom.xml
```diff
@@ -61,6 +61,13 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>${junit.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
```
rdiachenko, line 70: since you added bom, you should no longer need explicit versions in the below junit dependences, please remove `<version>${junit.version}</version>` from them, so they all rely on bom

rewrite.yml
MonuChaudhary14: This recipe requires some mandatory parameters

In the older version of OpenRewrite,  the plugin was was lenient. It silently ignored the misconfigured recipe and didn't crash.

In the newly bumped 6.42.0 version, the OpenRewrite team introduced strict recipe validation.

When the new plugin saw the InlineMethodCalls recipe missing its methodPattern, it threw a validation error and then crashed with a NullPointerException when trying to parse the missing pattern string.

MonuChaudhary14: https://docs.openrewrite.org/recipes/java/inlinemethodcalls#usage

pom.xml
MonuChaudhary14: done

rewrite.yml
rdiachenko: all right, let's revert this change, since you don't bump openrewrite version here, and properly fix it in the pr where we bump the version

MonuChaudhary14: okay

MonuChaudhary14: done

rdiachenko: close PR from dependabot and raise similar PR yourself with additional fixes. better to have fixes together with version bump in a single PR to keep context

MonuChaudhary14: should i make changes in this PR, or create a new individual one for every bump

rdiachenko (approved):
ok to merge after ci passes, but @MonuChaudhary14 please aim to do such fixes in the PRs where we update version. so if version update leads to issues, close dependabot PR and raise your own PR with version bump + fixes specifically for this version bump if we need any

rewrite.yml
rdiachenko: https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/270#pullrequestreview-4569085894

MonuChaudhary14:
> ok to merge after ci passes, but @MonuChaudhary14 please aim to do such fixes in the PRs where we update version. so if version update leads to issues, close dependabot PR and raise your own PR with version bump + fixes specifically for this version bump if we need any

okay, i will remember that

Issue #275, opened 2026-06-26
NullPointerException stack trace printed during execution when recipe validation fails (missing InlineMethodCalls parameters)


MonuChaudhary14:
After the merge of this PR https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/255

In older versions of OpenRewrite (rewrite-maven-plugin 6.17.0 and rewrite-rewrite 0.12.1), the plugin was lenient. If a recipe like org.openrewrite.java.InlineMethodCalls was included in a composite recipe (rewrite.yml) without its mandatory parameters (methodPattern and replacement), the plugin silently ignored the misconfiguration without any console spam.

After upgrading to rewrite-maven-plugin 6.42.0 and rewrite-rewrite 0.26.1, the OpenRewrite team introduced strict recipe validation.

When the new plugin encounters the InlineMethodCalls recipe missing its methodPattern, it correctly logs a validation error. However, instead of skipping the invalid recipe, the plugin logs Execution will continue regardless and attempts to execute it anyway. This results in MethodMatcher throwing a NullPointerException when trying to parse the missing pattern string.

The framework catches this exception and prints the entire stack trace to the console. The build does not crash and ultimately finishes with BUILD SUCCESS, but the noisy exception is confusing and alarming for users.

```
[INFO] Using active styles(s) []
[INFO] Validating active recipes...
[ERROR] Recipe validation error in org.openrewrite.config.CompositeRecipe for property org.openrewrite.java.InlineMethodCalls.methodPattern: is required
[ERROR] Recipe validation error in org.openrewrite.config.CompositeRecipe for property org.openrewrite.java.InlineMethodCalls.replacement: is required
[ERROR] Recipe validation error in org.openrewrite.config.CompositeRecipe for property org.openrewrite.java.InlineMethodCalls.methodPattern: is required
[ERROR] Recipe validation error in org.openrewrite.config.CompositeRecipe for property org.openrewrite.java.InlineMethodCalls.replacement: is required
[ERROR] Recipe validation errors detected as part of one or more activeRecipe(s). Execution will continue regardless.
[INFO] Project [checkstyle openrewrite recipes] Resolving Poms...
[INFO] Project [checkstyle openrewrite recipes] Parsing source files
[INFO] Running recipe(s)...
[WARNING] The recipe produced 182 warning(s). Please report this to the recipe author.
[...]

PR #276, opened 2026-06-26
Issue #275: Fixed the NullPointerException due to missing InlineMethodCalls parameters

MonuChaudhary14:
Issue #275

### DIFF
--- a/rewrite.yml
+++ b/rewrite.yml
@@ -51,7 +51,7 @@ recipeList:
   - org.openrewrite.java.recipes.NoMutableStaticFieldsInRecipes
   - org.openrewrite.java.recipes.RecipeEqualsAndHashCodeCallSuper
   - org.openrewrite.java.recipes.UseTreeRandomId
-  - org.openrewrite.java.InlineMethodCalls
+  - org.openrewrite.recipes.rewrite.InlineMethods
   - org.openrewrite.staticanalysis.NeedBraces
   - org.openrewrite.staticanalysis.RemoveSystemOutPrintln
   - org.openrewrite.java.RemoveAnnotation:


rewrite.yml
```diff
@@ -51,7 +51,7 @@ recipeList:
   - org.openrewrite.java.recipes.NoMutableStaticFieldsInRecipes
   - org.openrewrite.java.recipes.RecipeEqualsAndHashCodeCallSuper
   - org.openrewrite.java.recipes.UseTreeRandomId
-  - org.openrewrite.java.InlineMethodCalls
+#  - org.openrewrite.java.InlineMethodCalls
```
rdiachenko: please find a way to fix it and configure this recipe with needed parameters instead of just disabling it

MonuChaudhary14: The `InlineMethodCalls` recipe is a `find-and-replace` template recipe, and it requires the `methodPattern` and `replacement` parameters to work. To configure this, we need to define some target methods with their replacements. Alternatively, we could add a dummy method just to silence the validation error

A better option might be to replace it with `org.openrewrite.recipes.rewrite.InlineMethods`
This is a composite recipe (https://docs.openrewrite.org/recipes/recipes/rewrite/inlinemethods) that comes pre-packed with multiple configured recipes

Let me know which direction you'd prefer

rdiachenko: I'm good with the second option. could you update to `org.openrewrite.recipes.rewrite.InlineMethods` and see what we get?

MonuChaudhary14: done

rdiachenko (approved):
ok to merge after ci passes

Issue #288, opened 2026-06-30
Cover all pitest mutations for UseEnhancedSwitch


MonuChaudhary14:
Remove the suppressions block of UseEnhancedSwitch from pitest-recipe-2-suppressions.xml 
https://github.com/checkstyle/checkstyle-openrewrite-recipes/blob/main/config/pitest-recipe-2-suppressions.xml

PR #289, opened 2026-06-30
Issue #288: Removed some suppressions block of UseEnhancedSwitch

MonuChaudhary14:
Issue #288

### DIFF
--- a/src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
+++ b/src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
@@ -627,21 +627,12 @@ public class UseEnhancedSwitch extends Recipe {
             }
 
             @Override
-            public J.Return visitReturn(J.Return returnObj, Void voidObj) {
-                hasDisallowedFlow = true;
-                return returnObj;
-            }
-
-            @Override
-            public J.Break visitBreak(J.Break breakObj, Void voidObj) {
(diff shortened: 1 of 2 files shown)


MonuChaudhary14:
@sabaka @rdiachenko 

Please review this

src/main/java/org/checkstyle/autofix/recipe/UseEnhancedSwitch.java
```diff
@@ -627,21 +628,21 @@ public boolean hasDisallowedFlow() {
             }
 
             @Override
-            public J.Return visitReturn(J.Return returnObj, Void voidObj) {
+            public J.@Nullable Return visitReturn(J.Return returnObj, Void voidObj) {
                 hasDisallowedFlow = true;
-                return returnObj;
+                return null;
```
sabaka: Returning null is a strange fix. I think JavaVisitor is nullmarked and return of this method isn't marked as nullable in the parent class. This can potentially lead to a bug if the parent class changes. Could you explain why it is done this way please?

MonuChaudhary14: You are right, this is risky. Sorry, I missed that! 
The changes were originally made to kill the mutations generated by NullReturnValsMutator
I have found a different method to kill those mutations
Please check the new changes

sabaka (approved):
This looks better

Issue #245, opened 2026-06-15
Add support for ordering constructors by increasing parameter count in ConstructorDeclarationGrouping Recipe


lithops-zty:
`ConstructorDeclarationGrouping` recently gets a new property `orderByIncreasingParameterCount` after this PR: https://github.com/checkstyle/checkstyle/pull/19818

If enabled, we should make sure after autofix, the final constructor ordering is by increasing number of parameters.

lithops-zty:
I'm on it.

lithops-zty:
I'd like to recieve opinion on how to deal with the following edge case:
```java
/*xml
<module name="Checker">
  <module name="TreeWalker">
    <module
        name="com.puppycrawl.tools.checkstyle.checks.coding.ConstructorsDeclarationGroupingCheck">
      <property name="orderByIncreasingParameterCount" value="true"/>
    </module>
    <module name="SuppressionXpathSingleFilter">
      <property name="checks" value="ConstructorsDeclarationGroupingCheck"/>
      <property name="message" value="Constructors should be grouped together.*"/>
      <property name="query"
          value="//CTOR_DEF[not(./PARAMETERS/PARAMETER_DEF)
                or count(./PARAMETERS/PARAMETER_DEF)=3]"/>
    </module>
  </module>
</module>
*/

package org.checkstyle.autofix.recipe.constructorsdeclarationgrouping.orderwithgroupingviolationsuppressed;

public class InputOrderWithGroupingViolationSuppressed {

    InputOrderWithGroupingViolationSuppressed(int x) {}

    InputOrderWithGroupingViolationSuppressed(int x, boolean b) {}

    void foo() {}

    InputOrderWithGroupingViolationSuppressed(String s, int x, int y) {}

    InputOrderWithGroupingViolationSuppressed() {}
    // violation above 'Constructors should be ordered.*'
}
```
In words, the 0-arg and 3-arg constructors have grouping violations but are suppressed.
This leave only one ordering violation on 0-arg ctor.

What should the intended output be? Problem is we can't fix ordering without fixing grouping first. The suppression and violation cannot be respected simultaneously.

lithops-zty:
My current implementation disregards the suppression and groups (then sorts) the 0-arg ctor regardless:

```java
/*xml
<module name="Checker">
  <module name="TreeWalker">
    <module
        name="com.puppycrawl.tools.checkstyle.checks.coding.ConstructorsDeclarationGroupingCheck">
[...]

Issue #236, opened 2026-06-11
Kill the equivalent PIT mutations which require another type of marker


MonuChaudhary14:
### Background
During PIT mutation testing, equivalent mutations survive in the marker filtering logic across various recipes (e.g., `FinalClass` and `FinalLocalVariable`).

**Surviving Mutations:**
```xml
  <mutation unstable="false">
    <sourceFile>FinalClass.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.recipe.FinalClass$FinalClassVisitor</mutatedClass>
    <mutatedMethod>lambda$visitClassDeclaration$0</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.BooleanTrueReturnValsMutator</mutator>
    <description>replaced boolean return with true for org/checkstyle/autofix/recipe/FinalClass$FinalClassVisitor::lambda$visitClassDeclaration$0</description>
    <lineContent>.anyMatch(marker -&gt; marker.isFor(CheckFullName.FINAL_CLASS));</lineContent>
  </mutation>
  
  <mutation unstable="false">
    <sourceFile>FinalLocalVariable.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.recipe.FinalLocalVariable$LocalVariableVisitor</mutatedClass>
    <mutatedMethod>lambda$isViolationMarked$0</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.BooleanTrueReturnValsMutator</mutator>
    <description>replaced boolean return with true for org/checkstyle/autofix/recipe/FinalLocalVariable$LocalVariableVisitor::lambda$isViolationMarked$0</description>
    <lineContent>.anyMatch(marker -&gt; marker.isFor(CheckFullName.FINAL_LOCAL_VARIABLE));</lineContent>
  </mutation>
```
These mutations originate from lines similar to:
```
final boolean hasMarker = result.getMarkers()
        .findAll(CheckstyleViolationMarker.class).stream()
        .anyMatch(marker -> marker.isFor(CheckFullName.FINAL_CLASS));
```

Why it survives
PIT targets the lambda function marker -> marker.isFor(...). Because this lambda returns a boolean, PIT applies the BooleanTrueReturnValsMutator. This modifies the compiled code to ignore the actual condition and simply evaluate as: marker -> true.

[...]

PR #292, opened 2026-07-01
Issue #245: Implement orderByIncreasingParameterCount property support for ConstructorsDeclarationGrouping recipe

lithops-zty:
Issue: #245

### DIFF
--- a/config/pitest-recipe-2-suppressions.xml
+++ b/config/pitest-recipe-2-suppressions.xml
@@ -14,46 +14,30 @@
   <mutation unstable="false">
     <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
     <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
-    <mutatedMethod>lambda$isViolationMarked$0</mutatedMethod>
+    <mutatedMethod>hasViolationMarkerMatching</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.experimental.NakedReceiverMutator</mutator>
+    <description>replaced call to java/util/stream/Stream::filter with receiver</description>
+    <lineContent>.filter(marker -&gt; marker.isFor(targetCheck))</lineContent>
+  </mutation>
+
+  <!-- This boolean method is currently equivalent to hasViolationMarker until the check has a third violation type -->
+  <mutation unstable="false">
+    <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
+    <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
+    <mutatedMethod>lambda$hasGroupingViolationMarker$0</mutatedMethod>
     <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.BooleanTrueReturnValsMutator</mutator>
-    <description>replaced boolean return with true for org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor::lambda$isViolationMarked$0</description>
-    <lineContent>.anyMatch(marker -&gt; marker.isFor(targetCheck));</lineContent>
+    <description>replaced boolean return with true for org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor::lambda$hasGroupingViolationMarker$0</description>
+    <lineContent>return marker.violation().getMessage()</lineContent>
   </mutation>
 
+  <!-- Cannot be killed by integration tests until a second recipe lands a marker on this node type. See issue #236 -->
   <mutation unstable="false">
     <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
     <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
-    <mutatedMethod>reorderConstructors</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_ORDER_IF</mutator>
-    <description>removed conditional - replaced comparison check with true</description>
-    <lineContent>while (insertIndex &lt; statements.size()</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
-    <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
-    <mutatedMethod>visitClassDeclaration</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.NonVoidMethodCallMutator</mutator>
-    <description>removed call to java/util/List::isEmpty</description>
-    <lineContent>if (firstConstructorIndex == -1 || violatingConstructors.isEmpty()) {</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
-    <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
-    <mutatedMethod>visitClassDeclaration</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_ELSE</mutator>
-    <description>removed conditional - replaced equality check with false</description>
-    <lineContent>if (firstConstructorIndex == -1 || violatingConstructors.isEmpty()) {</lineContent>
-  </mutation>
-
-  <mutation unstable="false">
-    <sourceFile>ConstructorsDeclarationGrouping.java</sourceFile>
-    <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
-    <mutatedMethod>visitClassDeclaration</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
-    <description>removed conditional - replaced equality check with true</description>
-    <lineContent>if (firstConstructorIndex == -1 || violatingConstructors.isEmpty()) {</lineContent>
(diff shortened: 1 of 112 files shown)


romani:
@MonuChaudhary14 , please review

src/main/java/org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping.java
```diff
@@ -64,25 +81,147 @@ public J.ClassDeclaration visitClassDeclaration(
...
             final J.ClassDeclaration newClassDeclaration;
-
-            // keep code intact if there are no constructors or no violations
-            if (firstConstructorIndex == -1 || violatingConstructors.isEmpty()) {
-                newClassDeclaration = classDeclaration;
-            }
-            else {
-                final List<Statement> reordered =
-                    reorderConstructors(statements, firstConstructorIndex, violatingConstructors);
-
```
MonuChaudhary14, line 79: There is a regression where the recipe modifies files even when Checkstyle emits 0 violations (if the user explicitly suppresses the rule for the entire class)

In the previous implementation, there was an early exit check in visitClassDeclaration that prevented modifying the AST if there were no violation markers. In the new implementation, this early exit was removed

When orderByIncreasingParameterCount is true, findConstructorsToSort() unconditionally adds the entire initial group to the candidates list and sorts them. If the recipe is activated by a violation in one class, it will visit every other class in the file. If it visits a completely suppressed class (0 violation markers), the recipe still sorts the initial group and rewrites the file, ignoring the suppression

Add these input and output file also:
Input:
```
/*xml
<module name="Checker">
  <module name="TreeWalker">
    <module
        name="com.puppycrawl.tools.checkstyle.checks.coding.ConstructorsDeclarationGroupingCheck">
      <property name="orderByIncreasingParameterCount" value="true"/>
    </module>
  </module>
  <module name="SuppressionXpathSingleFilter">
    <property name="checks" value="ConstructorsDeclarationGrouping"/>
    <property name="query" value="//CLASS_DEF[./IDENT[@text='NestedClass']]//CTOR_DEF"/>
  </module>
</module>
*/

package org.checkstyle.autofix.recipe.constructorsdeclarationgrouping.ordernoviolationoutoforder;

public class InputOrderNoViolationOutOfOrder {

    InputOrderNoViolationOutOfOrder(int x) {}

    InputOrderNoViolationOutOfOrder() {} // violation 'Constructors should be ordered by increasing parameter count.'

    static class NestedClass {
        NestedClass(int x) {}
        NestedClass() {}
    }
}
```
Output:
```
/*xml
<module name="Checker">
  <module name="TreeWalker">
    <module
        name="com.puppycrawl.tools.checkstyle.checks.coding.ConstructorsDeclarationGroupingCheck">
      <property name="orderByIncreasingParameterCount" value="true"/>
    </module>
  </module>
  <module name="SuppressionXpathSingleFilter">
    <property name="checks" value="ConstructorsDeclarationGrouping"/>
    <property name="query" value="//CLASS_DEF[./IDENT[@text='NestedClass']]//CTOR_DEF"/>
  </module>
</module>
*/

package org.checkstyle.autofix.recipe.constructorsdeclarationgrouping.ordernoviolationoutoforder;

public class OutputOrderNoViolationOutOfOrder {

    OutputOrderNoViolationOutOfOrder() {}

    OutputOrderNoViolationOutOfOrder(int x) {}

    static class NestedClass {
        NestedClass(int x) {}
        NestedClass() {}
    }
}
```

MonuChaudhary14 (changes requested):
Please fix this

src/main/java/org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping.java
lithops-zty: @MonuChaudhary14 Thanks for the catch and the test case.

The regression was due to findConstructorsToSort() unconditionally added the entire initial constructor group to the sort candidates. 

Fix: We now only add **sorted** constructors from the initial group to the sort candidates.

While fixing this, I found another bug: a constructor with grouping violation and suppressed ordering violation is incorrectly sorted:
```java
/*xml
<module name="Checker">
  <module name="TreeWalker">
    <module
        name="com.puppycrawl.tools.checkstyle.checks.coding.ConstructorsDeclarationGroupingCheck">
      <property name="orderByIncreasingParameterCount" value="true"/>
    </module>
    <module name="SuppressionXpathSingleFilter">
      <property name="checks" value="ConstructorsDeclarationGrouping"/>
      <property name="message" value="Constructors should be ordered.*"/>
      <property name="query"
          value="//CTOR_DEF[./PARAMETERS/PARAMETER_DEF/TYPE/IDENT]"/>
    </module>
  </module>
</module>
*/

package org.checkstyle.autofix.recipe.constructorsdeclarationgrouping.orderwithgroupingonly;

public class InputOrderWithGroupingOnly {

    InputOrderWithGroupingOnly() {}

    InputOrderWithGroupingOnly(boolean b, int x) {}

    void foo() {}

    InputOrderWithGroupingOnly(String s) {} // violation 'Constructors should be grouped together.*'
}
```
 Incorrect:
```java
/*xml
...
*/

package org.checkstyle.autofix.recipe.constructorsdeclarationgrouping.orderwithgroupingonly;

public class OutputOrderWithGroupingOnly {

    OutputOrderWithGroupingOnly() {}

    OutputOrderWithGroupingOnly(String s) {}

    OutputOrderWithGroupingOnly(boolean b, int x) {}

    void foo() {}
}
```

Correct:
```java
/*xml
...
*/

package org.checkstyle.autofix.recipe.constructorsdeclarationgrouping.orderwithgroupingonly;

public class OutputOrderWithGroupingOnly {

    OutputOrderWithGroupingOnly() {}

    OutputOrderWithGroupingOnly(boolean b, int x) {}

    OutputOrderWithGroupingOnly(String s) {}

    void foo() {}
}
```
Fixed and testcase added.

config/pitest-recipe-2-suppressions.xml
```diff
@@ -10,53 +9,42 @@
...
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
-    <description>removed conditional - replaced equality check with true</description>
-    <lineContent>if (firstConstructorIndex == -1 || violatingConstructors.isEmpty()) {</lineContent>
+    <mutatedMethod>lambda$hasViolationMarkerMatching$2</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.BooleanTrueReturnValsMutator</mutator>
+    <description>replaced boolean return with true for org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor::lambda$hasViolationMarkerMatching$2</description>
+    <lineContent>.filter(marker -&gt; marker.isFor(targetCheck))</lineContent>
   </mutation>
 
-  <!-- Cannot be killed by integration tests until a second recipe lands a marker on this node type. See issue #236 -->
```
romani, line 59: @lithops-zty  , please restore all comment like this.

romani (changes requested):
items:

src/main/java/org/checkstyle/autofix/recipe/ConstructorsDeclarationGrouping.java
```diff
@@ -64,25 +87,176 @@ public J.ClassDeclaration visitClassDeclaration(
...
+
+        /**
+         * Returns the number of parameters of a constructor.
+         *
+         * @param method the constructor
+         * @return parameter count, 0 for a no-arg constructor
+         */
+        private static int countParameters(J.MethodDeclaration method) {
+            return method.getMethodType().getParameterTypes().size();
+        }
```
MonuChaudhary14, line 258: In this method, you are using `method.getMethodType().getParameterTypes().size()`, but it is at risk for a NullPointerException

In OpenRewrite, getMethodType() evaluates to null if the parser encounters unresolved types or missing dependencies. Calling .getParameterTypes() on this null value will throw a NullPointerException and crash the recipe mid-execution

Replace it with:
```
private static int countParameters(J.MethodDeclaration method) {
    final int parameterCount;
    if (method.getParameters().get(0) instanceof J.Empty) {
        parameterCount = 0;
    }
    else {
        parameterCount = method.getParameters().size();
    }
    return parameterCount;
}
```

MonuChaudhary14 (changes requested):
Changes

config/pitest-recipe-2-suppressions.xml
```diff
@@ -10,53 +9,42 @@
...
-    <mutatedClass>org.checkstyle.autofix.recipe.ConstructorsDeclarationGrouping$ConstructorsDeclarationGroupingVisitor</mutatedClass>
-    <mutatedMethod>reorderConstructors</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_ORDER_IF</mutator>
-    <description>removed conditional - replaced comparison check with true</description>
-    <lineContent>while (insertIndex &lt; statements.size()</lineContent>
+    <mutatedMethod>collectGroupedAndSortedConstructors</mutatedMethod>
+    <mutator>org.pitest.mutationtest.engine.gregor.mutators.RemoveConditionalMutator_EQUAL_IF</mutator>
+    <description>removed conditional - replaced equality check with true</description>
+    <lineContent>&amp;&amp; method.isConstructor()</lineContent>
   </mutation>
```
MonuChaudhary14: If possible, kill the mutations. If not, please state the reason they survived

src/test/java/org/checkstyle/autofix/recipe/ConstructorsDeclarationGroupingTest.java
```diff
@@ -270,4 +270,115 @@ void idempotent(ReportParser parser) throws Exception {
...
+
+    /**
+     * The last constructor is separated by a method and has its ordering
+     * violation suppressed. It should be grouped but not reordered — moving
+     * one position forward to join the constructor group.
+     */
+    @RecipeTest
+    void orderWithGroupingOnly(ReportParser parser) throws Exception {
+        verify(parser, "OrderWithGroupingOnly");
+    }
```
romani: comment on methods here is too far from where reader should consider it.
ConstructorsDeclarationGroupingTest.java is not class reviewer should read, it is just technical launchers.

you should move it to OrderWithGroupingOnly.java to let users understand coverage goal of this Input

romani (changes requested):
last:

src/test/java/org/checkstyle/autofix/recipe/ConstructorsDeclarationGroupingTest.java
lithops-zty: Makes sense! I'll move all comments to respective inputs

config/pitest-recipe-2-suppressions.xml
lithops-zty: Killed one; out of three remaining:
- Two are related to issue #236;
- One is due to `hasGroupingViolationMarker` being equivalent to `hasViolationMarker` when `hasOrderingViolationMarker` is false. This will not hold after a third violation type is introduced.

romani (approved):
thanks a lot.

very good step forward.

MonuChaudhary14 (approved):
lgtm

PR #297, opened 2026-07-04

MonuChaudhary14:
Issue #237


MonuChaudhary14:
@rdiachenko @sabaka 

Please review this PR

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -197,8 +196,7 @@ private UUID findEnclosingNode(CheckstyleViolation violation,
         }
 
         private boolean enclosesLine(Range range, CheckstyleViolation violation) {
```
sabaka: Naming reads wrong now, you no longer validate if the violation is enclosed by a given range.

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -153,32 +155,29 @@ private void processFileViolations(J.CompilationUnit compilationUnit, Path sourc
...
-
-            final Map<UUID, List<CheckstyleViolationMarker>> result = new HashMap<>();
-            if (smallestNodeId != null) {
-                final UUID targetId = resolveTargetNode(smallestNodeId, violation,
-                        parentMap, nodeTrees);
-                result.computeIfAbsent(targetId, nodeId -> new ArrayList<>())
-                        .add(new CheckstyleViolationMarker(Tree.randomId(), violation));
-            }
-            return result;
+            return Stream.of(false, true)
```
sabaka: `Stream.of(findEnclosingNode(violation, nodeRanges, false), findEnclosingNode(violation, nodeRanges, true))` would read better. You don't care much about `true` and `false` here, but care about node IDs.

Would be even better if you extract ids search to a method returning `Optional<UUID>` and use it as a start of the chain instead of creating stream.
Something like

```java
private Optional<UUID> findEnclisingNodeId(CheckstyleViolation violation, Map<UUID, Range> nodeRanges) {
 return Optiona.ofNullable(findEnclosingNode(violation, nodeRanges, false))
   .or(() -> findEnclosingNode(violation, nodeRanges, true));
}
```

and then
```java
return findEnclisingNodeId(violation, nodeRanges)
  .map(smallestNodeId -> ...)
  .orElseGet(...)
```

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

MonuChaudhary14:
@sabaka 

Please review this PR

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -223,37 +223,17 @@ private UUID resolveTargetNode(UUID startNodeId, CheckstyleViolation violation,
...
-                    }
-                }
-                if (range.endLine() == violation.getLine()) {
-                    if (range.endCol() >= violation.getColumn()) {
-                        result = true;
-                    }
-                }
-            }
-            return result;
+            return range.startLine() != range.endLine();
```
sabaka: You no longer use `violation` in this method.
Semantic has also changed dramatically, it only validates if `range` is multiline. There is just one usage, and it's a one-liner, maybe move it in inside `encloses` method.
Also validate if logic is correct as the condition looks strange without using validation.

MonuChaudhary14: In this PR, I have redesigned the logic for how node resolution works

In the previous version, inside `BoundingBoxCapture.apppend()`, the code was allocating a new `ArrayList`, walking the cursor path, and re-allocating Range objects to update the endLine and endCol for every single ancestor node on every single string chunk printed. This resulted in `O(Tree Depth)` allocations and map updates for every token in the file.

The previous encloses() method tried to strictly evaluate endLine and endCol across multi-line elements, making the logic too complex. When it failed, it fell back to `startsBeforeOrAtLine(),` which was buggy because it completely ignored columns, means it could accidentally attach a violation to a node that started on the same line but after the violation's column.

To fix this, I deleted the pathTrees loop in `BoundingBoxCapture.append()`. We no longer constantly update the end bounds of AST nodes. Instead, we only capture the startLine and startCol of each node exactly once in O(1) time via putIfAbsent().

Since we no longer track the bounding box area, I replaced the complex comparator. The new comparator sorts valid nodes in descending order by startLine and startCol ( .reversed() ). Instead of looking for the smallest bounding box, we now look for the node with the latest start position that occurs before or at the violation.

This is safe becuase a child node will always start at or after its parent. Therefore, the node immediately preceding a violation will always be on the exact same branch of the tree as the actual target node.

Even if it picks a deep descendant node (like a VariableDeclaration), the existing `resolveTargetNode() `logic correctly climbs the parentMap and resolve to the requested target type (like ClassDeclaration).

Mathematical Explannation:
Let N be the deepest node that contains the violation at position P , `N.start <= P` Any node X where `N.start <= X.start <= P` must be:

- A descendant of N (it starts within N's span and before P), orN itself. It cannot be a sibling or cousin of N, because siblings come either before or after N in the parent's child list:

- If X comes before N in tree order, then `X.start < N.start`, contradicts `X.start >= N.start`
- If X comes after N in tree order, then X starts after N ends, meaning `X.start > N.end >= P`, contradicts `X.start <= P`

Therefore, the `latest start ≤ P` node is always a descendant of N (or N itself). Walking up via `resolveTargetNode` from node will always reach N or N's ancestor of the correct type

```
class A {          // starts line 1
    int x = 1;     // starts line 2 (x ends here too)
}                  // A ends line 3
class B {          // starts line 4
    int y = 2;     // starts line 5, violation at (5, 9)
}
```
Nodes passing encloses for violation at (5, 9): `CompilationUnit, A, x, B, y,` all start before line 5. But the comparator (reversed) picks y `latest start: line 5`. resolveTargetNode walks up from y to the correct target. Correct result, even though A and x passed the filter.

There is a little bit of clean up remaining regarding the endLine/endCol fields in the Range record. I will handle this in the next PR, as those values are linked with the unit tests in ViolationMarkerRecipeTest

sabaka: Thanks for extensive comment.
I see you have changed
```
return range.startLine() != range.endLine();
```
that looked like a bug to me, to
```
result = range.startLine() < violation.getLine();
```
which makes sense. What surprises me is that no tests were failing. Do we have a test case touching this code and validating effects of the condition?

MonuChaudhary14: The old code was working because I had already deleted the expensive update loop in `BoundingBoxCapture` This meant that endLine was permanently 0 for every node. Because endLine was always 0, the condition range.startLine() != range.endLine() always evaluated to true.

Therefore, in the encloses() method, whenever range.startLine() < violation.getLine(), it called checkPartialLines(), which simply returned true

So, encloses() was already acting as a coarse filter that returned true for all nodes that started before or at the violation,which is exactly what the new code does

Once all those preceding nodes passed that filter, the new `.reversed()` comparator (which I also introduced) sorted them and correctly picked the node with the highest start coordinates.

The tests passed because the math accidentally worked out perfectly to implement the Latest Start Position

MonuChaudhary14:
@sabaka 

Please review this PR

sabaka (approved):

Issue #325, opened 2026-07-16
Ensure every Checkstyle config produces at least one violation on the Kafka project


MonuChaudhary14:
# Description
Implement a validation mechanism to verify that every Checkstyle configuration produces at least one violation when executed against the Kafka project(if it doesn't produce violations, then the recipe logic is not executed, so we don't really test it)

romani:
let be careful with automation of this, I would recommend to make this manually, make sure it works, be aligned on way it will be stored, and only at the end do automation.

MonuChaudhary14:
@romani @rdiachenko 

Issue is fixed, please close this

PR #329, opened 2026-07-17
Issue #325: Ensure every Checkstyle config produces at least one violation on the Kafka project

MonuChaudhary14:
Issue #325

### DIFF
--- a/.ci/generate-diff-report-for-kafka.sh
+++ b/.ci/generate-diff-report-for-kafka.sh
@@ -4,7 +4,7 @@ set -e
 RECIPES_DIR="$(cd "$(dirname "$0")/.." && pwd)"
 WORK_DIR="${RECIPES_DIR}/.ci-temp/diff-report-kafka"
 TARGET_REPO="https://github.com/apache/kafka.git"
-TARGET_BRANCH="trunk"
+TARGET_SHA="fdc59be8c84af1424b2f51a2df81386927f01a90"
 
 function getMavenProperty {
   property="\${$1}"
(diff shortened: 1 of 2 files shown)


MonuChaudhary14:
@romani 

Can you please tell me whether i understand this issue correct or not
Now, the script will now verify that the checkstyle-report.xml contains at least one <error> element before proceeding to run rewrite. If no violations are found, it will fail with an error message

.ci/generate-diff-report-for-kafka.sh
```diff
@@ -70,6 +70,13 @@ java -jar "$CHECKSTYLE_JAR" \
   "$WORK_DIR/kafka" \
   || true
 
+echo "Verifying that the Checkstyle configuration produced at least one violation..."
+if ! grep -q "<error " "$WORK_DIR/checkstyle-report.xml"; then
+    echo "ERROR: No Checkstyle violations found in the Kafka project!"
```
romani: I think we need to make it more complicated.
We need to check that each module in checktyle config has at least 1 violation in report

romani (changes requested):
item:

.ci/generate-diff-report-for-kafka.sh
MonuChaudhary14: Sorry, but I still don't get it. What if there isn't any violation for that specific module in the repo? What should we do in that case?

romani: find SHA where we have violation. 
if we not do this, some recipe might never run as it has no violation to deal we will think that "ALL" recipies executed and worked as before. It will be very bad illusion. 

if we can not find such SHA to satisfy all it will complicate situation.

MonuChaudhary14: I was working on the exclusion file list, and i found out that the crash is happening due to our recipe, i am still reseaching more about it, if we can solve this, then maybe we can avoid SHA and work on latest code

And if any crash found happens in future, then it maybe our code fault

MonuChaudhary14: > find SHA where we have violation. if we not do this, some recipe might never run as it has no violation to deal we will think that "ALL" recipies executed and worked as before. It will be very bad illusion.
> 
> if we can not find such SHA to satisfy all it will complicate situation.

Thank you,now i got it, i will look for SHA which satisfy the condition

rdiachenko: Now we rely on default check configuration. Another option would be to configure the check, set some properties (if the check has any) to non-default values that produce violations.

romani: yes, we can use non default to make to likely to have violation.

MonuChaudhary14: Thank you, I got it. I’ll try it.

I’m currently busy with some work and will be available on Sunday after 6 PM. I hope you understand.

MonuChaudhary14: I have checked multiple SHA, but we have 4 modules in which there are no violation and also they don't have properties

EmptyStatement
RedundantImport
UpperEll

and one is AvoidStarImport, whose properties help in relaxing the code, 
These 4 modules produces no violation

romani: put them in some sort of suppression list , lets move forward with merge and enforcing something that we can do now, we will improve on others later on.

MonuChaudhary14: Sorry for the delay, please check this
This SHA needs 5 modules to be suppressed ,also check this issue https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/330

romani (approved):
let move forward

rdiachenko (approved):
lgtm

Issue #339, opened 2026-07-24
UseEnhancedSwitch incorrectly converts switch statements to return switch expressions


MonuChaudhary14:
# Description

The UseEnhancedSwitch recipe incorrectly converts exhaustive switch statements into return switch expressions even when all case branches end in a throw statement. This results in invalid Java code being generated, causing a switch expression does not have any result expressions compiler error.
Input
```
if (simulateError.hasError) {
    switch (simulateError.errorType) {
        case IO_EXCEPTION:
            throw new IOException("Simulated IO error");
        default:
            throw new RuntimeException("Simulated Runtime error");
    }
}
return new UnifiedLog(...);
```

Current Output:
```
if (simulateError.hasError) {
    return switch (simulateError.errorType) {
        case IO_EXCEPTION -> throw new IOException("Simulated IO error");
        default -> throw new RuntimeException("Simulated Runtime error");
    }; // Compiler Error: switch expression does not have any result expressions
}
return new UnifiedLog(...);
```

Expected Behaviour:
```
if (simulateError.hasError) {
    switch (simulateError.errorType) {
        case IO_EXCEPTION -> throw new IOException("Simulated IO error");
        default -> throw new RuntimeException("Simulated Runtime error");
    }
}
return new UnifiedLog(...);

```

romani:
@MonuChaudhary14 , in what source code did you find it ?

MonuChaudhary14:
> [@MonuChaudhary14](https://github.com/MonuChaudhary14) , in what source code did you find it ?

https://github.com/apache/kafka/blob/2699fc40748da5ce9dc4b6aac41687fbc603ab1d/storage/src/test/java/org/apache/kafka/storage/internals/log/LogLoaderTest.java#L268

MonuChaudhary14:
This is the reason, the previous CI of silent failure PR failure (https://github.com/checkstyle/checkstyle-openrewrite-recipes/actions/runs/30040901539/job/89320272012)

romani:
fix is merged

PR #340, opened 2026-07-24
Issue #339: Fixed the enhace switch return error

MonuChaudhary14:
Issue #339

### DIFF
--- a/.ci/kafka-init.gradle.template
+++ b/.ci/kafka-init.gradle.template
@@ -19,8 +19,6 @@ rootProject {
         configFile = file("@WORK_DIR@/resolved-diff-report-rewrite.yml")
 
         exclusion(
-            //until https://github.com/checkstyle/checkstyle-openrewrite-recipes/issues/339
-            "**/LogLoaderTest.java"
         )
     }
     dependencies {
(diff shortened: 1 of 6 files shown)


romani:
@MonuChaudhary14 , please rebase and remove exclude , and we should see it in diff report.

MonuChaudhary14:
> @MonuChaudhary14 , please rebase and remove exclude , and we should see it in diff report.

done

romani:
Off topic 

From diff report https://github.com/checkstyle/checkstyle-openrewrite-recipes/actions/runs/30115157400?pr=340

Why our recipe put Test annotation on same line as method ?

likely due to craze setting to get violations, but this make big noise, such report is hard to review, single file cause tons of violations.

I think we naturally came to point that we need report for single recipe.

@MonuChaudhary14 , do you see easy way to grep from git diff what recipes are changes and keep only it in config ?

MonuChaudhary14:
> Off topic
> From diff report https://github.com/checkstyle/checkstyle-openrewrite-recipes/actions/runs/30115157400?pr=340

In the main branch, the LogLoaderTest.java, file is in exclusion, but in this PR, it is removed so it shows the whole diff of the file

> Why our recipe put Test annotation on same line as method ?

 The checkstyle throws violation in that file, so the recipe(AnnotationOnSameLine) put the Test to same line

> I think we naturally came to point that we need report for single recipe.

Let me see, what we can do

rdiachenko (approved):
lgtm

MonuChaudhary14:
@romani 

I dont get it what you mean by running only one recipe
Becuase if a recipe is changing, then we only see the diff or that recipe, because it compare it with main diff

Do you want we should only run those recipe that we change in PR, but it will only decrease the script executon time, but i think the diff will be same

romani:
I was wrong, diff was full only due to fact that whole file was excluded, whole file for all recipies, that will not be frequent for us I hope.

romani (approved):

PR #341, opened 2026-07-24

MonuChaudhary14:
Issue #237 

I have removed the redundant if (compUnit.getMarkers().findFirst(MarkersApplied.class).isPresent()) check from ScannerVisitor.visitCompilationUnit() in ViolationMarkerRecipe.java

In the OpenRewrite ScanningRecipe lifecycle, the MarkerVisitor already properly checks for the presence of the MarkersApplied marker before applying any changes to the AST

Because MarkerVisitor already guards against re-evaluating nodes where the markers have been applied, the check inside ScannerVisitor was a redundant optimization


rdiachenko (approved):
lgtm

PR #342, opened 2026-07-25

MonuChaudhary14:
Issue #237

### DIFF
--- a/src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
+++ b/src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
@@ -122,19 +122,19 @@ public class ViolationMarkerRecipe extends ScanningRecipe<Accumulator> {
                                            List<CheckstyleViolation> fileViolations) {
             final Map<UUID, Range> nodeRanges = new LinkedHashMap<>();
             final Map<UUID, UUID> parentMap = new HashMap<>();
-            final Map<UUID, Tree> nodeTrees = new HashMap<>();
+            final Map<UUID, Tree> treeNodes = new HashMap<>();
             final Cursor rootCursor = new Cursor(null, "root");
             final TreeVisitor<?, PrintOutputCapture<TreeVisitor<?, ?>>> printer =
                     compilationUnit.printer(rootCursor);
 
             final BoundingBoxCapture capture = new BoundingBoxCapture(
-                    printer, nodeRanges, parentMap, nodeTrees);
-            printer.visit(compilationUnit, capture, rootCursor);
+                    printer, nodeRanges, parentMap, treeNodes);
+            printer.visit(compilationUnit, capture.append(""), rootCursor);
 
             final Map<UUID, List<CheckstyleViolationMarker>> markersToAdd = new HashMap<>();
             for (CheckstyleViolation violation : fileViolations) {
                 final Map<UUID, List<CheckstyleViolationMarker>> result = findSmallestNode(
-                        violation, nodeRanges, parentMap, nodeTrees);
+                        violation, nodeRanges, parentMap, treeNodes);
                 for (Map.Entry<UUID, List<CheckstyleViolationMarker>> entry : result.entrySet()) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -129,7 +129,8 @@ private void processFileViolations(J.CompilationUnit compilationUnit, Path sourc
 
             final BoundingBoxCapture capture = new BoundingBoxCapture(
                     printer, nodeRanges, parentMap, nodeTrees);
-            printer.visit(compilationUnit, capture, rootCursor);
+            final PrintOutputCapture<TreeVisitor<?, ?>> capture2 = capture.append("");
```
rdiachenko: we don't need a new variable for this, please simplify

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -263,28 +264,28 @@ private void advance(String segment) {
         @Override
         public PrintOutputCapture<TreeVisitor<?, ?>> append(String text) {
             advance(out.substring(lastAppendEnd));
+            lastAppendEnd = out.length();
```
rdiachenko, line 266: can we move it closer to its usage? there is nothing in the below while loop that changes it

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -263,28 +264,28 @@ private void advance(String segment) {
...
                     nodeRanges.putIfAbsent(currentId,
                             new Range(line, column, 0, 0));
                     nodeTrees.putIfAbsent(currentId, treeNode);
-                    if (childId != null) {
-                        parentMap.putIfAbsent(childId, currentId);
+                    if (childTree != null) {
+                        parentMap.putIfAbsent(childTree.getId(), currentId);
                     }
-                    childId = currentId;
+                    childTree = treeNode;
```
rdiachenko: bad and inconsistent naming for both. does `treeNode` -> `nextNode` and `childTree` -> `node` make more sense? same for `currentId`. also `nodeTrees` doesn't make sense, it is a tree node but not the other way around

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -263,28 +264,28 @@ private void advance(String segment) {
         @Override
         public PrintOutputCapture<TreeVisitor<?, ?>> append(String text) {
             advance(out.substring(lastAppendEnd));
+            lastAppendEnd = out.length();
 
             final Cursor currentCursor = getContext().getCursor();
             final Iterator<Object> it = currentCursor.getPath();
-
-            UUID childId = null;
+            Tree childTree = null;
```
rdiachenko: i don't see we use it outside the `while` loop. let's replace `while` with `for` and move it inside as initial state

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
MonuChaudhary14: also, i have update nodeTrees in all the places, to make it meaningful, if you want i can revert it

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
MonuChaudhary14: The lastappendnd is used just above it

rdiachenko (approved):
lgtm

Issue #343, opened 2026-07-25
Optimize CI Diff Report Generation by filtering unchanged recipes


MonuChaudhary14:
# Description
Currently, our `.ci/generate-diff-report-for-checkstyle.sh` and `.ci/generate-diff-report-for-kafka.sh` scripts run Checkstyle and open-rewrite for all recipes against the target repositories (checkstyle and kafka) on every PR.

This process takes a significant amount of time and CI resources. Since unchanged recipes produce the exact same output on the baseline and the PR branches, they cancel each other out and result in zero diffs in the final HTML report. Running these unchanged recipes is computationally expensive and adds massive overhead to the PR build time.

MonuChaudhary14:
@rdiachenko 

This issue is also fixed

PR #344, opened 2026-07-25
Issue #343: Optimized CI Diff Report Generation by filtering unchanged recipes

MonuChaudhary14:
Issue #343

### DIFF
--- a/.ci/generate-diff-report-for-checkstyle.sh
+++ b/.ci/generate-diff-report-for-checkstyle.sh
@@ -28,6 +28,38 @@ cp "$RECIPES_DIR/.ci/diff-report-rewrite.yml" "$WORK_DIR/diff-report-rewrite.yml
 echo "Fetching base branch $BASE_BRANCH..."
 git fetch origin "$BASE_BRANCH"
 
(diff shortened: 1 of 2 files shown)


romani:
@rdiachenko , @sabaka , please share your thought on this ?

rdiachenko:
> @rdiachenko , @sabaka , please share your thought on this ?

there is a tradeoff here:

1. run all recipes and catch cases when new recipe fixes produce new violations that can be fixed by existing recipes (but we need to run checkstyle multiple times to catch new violations and re-run recipes, which is not the case right now)

or

2. apply this optimisation and filter out unchanged recipes

Since we don't have full support for 1 (we run checkstyle once and fix all that's applicable), and, realistically, i don't see 1 as a common use case. it is more like an exceptional case that is good to have covered later if we need it. I'm good to go with 2

sabaka:
I think running only changed recipes is good enough – our goal here is to catch regression introduced by changes in the recipe now. We should probably run all in cases when we change the core part affecting multiple recipes, but this can be done in the future. 
On a separate note, changes in both files are identical, should probably aim to extract it to the shell file as a function anld load it in both.

rdiachenko (approved):
lgtm

Issue #345, opened 2026-07-26
Implement Recipe for MissingSwitchDefault


MonuChaudhary14:
# Description

Check: https://checkstyle.sourceforge.io/checks/coding/missingswitchdefault.html

Check core OpenRewrite repository and reuse existing recipes if present.

Update README with the recipe link and notes about coverage if needed.

rdiachenko:
fix is merged

PR #346, opened 2026-07-26
Issue #345: Implemented the recipe for MissingSwitchDefault

MonuChaudhary14:
Issue #345

### DIFF
--- a/src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
+++ b/src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
@@ -1,0 +1,203 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+// checkstyle-openrewrite-recipes: Automatically fix Checkstyle violations with OpenRewrite.
+// Copyright (C) 2025 The Checkstyle OpenRewrite Recipes Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+///////////////////////////////////////////////////////////////////////////////////////////////
+
+package org.checkstyle.autofix.recipe;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+import org.checkstyle.autofix.CheckFullName;
+import org.checkstyle.autofix.marker.CheckstyleViolationMarker;
+import org.openrewrite.ExecutionContext;
+import org.openrewrite.Recipe;
+import org.openrewrite.Tree;
+import org.openrewrite.TreeVisitor;
+import org.openrewrite.java.JavaIsoVisitor;
+import org.openrewrite.java.tree.J;
+import org.openrewrite.java.tree.JContainer;
+import org.openrewrite.java.tree.JRightPadded;
+import org.openrewrite.java.tree.Space;
+import org.openrewrite.java.tree.Statement;
+import org.openrewrite.marker.Markers;
+
+public class MissingSwitchDefault extends Recipe {
+
+    private static final String NEW_LINE = "\n";
+    private static final String SPACE_STRING = " ";
+    private static final String DEFAULT_STRING = "default";
+
+    private static final Set<Class<?>> TERMINAL_STATEMENTS =
+            new HashSet<>(Arrays.asList(
+                    J.Break.class, J.Return.class, J.Throw.class, J.Continue.class
+            ));
+
+    public MissingSwitchDefault() {
+    }
+
+    @Override
+    public String getDisplayName() {
+        return "MissingSwitchDefault recipe";
+    }
+
+    @Override
+    public String getDescription() {
+        return "Adds a default case to switch statements if it is missing.";
+    }
+
+    @Override
+    public TreeVisitor<?, ExecutionContext> getVisitor() {
+        return new MissingSwitchDefaultVisitor();
+    }
+
+    private final class MissingSwitchDefaultVisitor extends JavaIsoVisitor<ExecutionContext> {
+
+        private MissingSwitchDefaultVisitor() {
+        }
+
+        @Override
+        public J.Switch visitSwitch(J.Switch switchStmt, ExecutionContext executionContext) {
+            J.Switch visited = super.visitSwitch(switchStmt, executionContext);
+
+            if (isViolationMarked(visited)) {
+                final J.Block casesBlock = visited.getCases();
+                List<Statement> cases = new ArrayList<>(casesBlock.getStatements());
+
+                final boolean isLambdaStyle = isLambdaStyleSwitch(cases);
+
+                if (!isLambdaStyle) {
+                    cases = tryAddBreakToLastCase(cases);
+                }
+
+                final J.Identifier defaultLabel = new J.Identifier(
+                        Tree.randomId(),
+                        Space.EMPTY,
+                        Markers.EMPTY,
+                        null,
+                        DEFAULT_STRING,
+                        null,
+                        null);
+
+                final JRightPadded<J> paddedLabel;
+                final JContainer<Statement> statements;
+                final J.Case.Type caseType;
+
+                if (isLambdaStyle) {
+                    paddedLabel = JRightPadded.build((J) defaultLabel)
+                            .withAfter(Space.build(SPACE_STRING, Collections.emptyList()));
+
+                    final J.Block emptyBlock = J.Block.createEmptyBlock()
+                            .withPrefix(Space.build(SPACE_STRING, Collections.emptyList()));
+                    statements = JContainer.build(
+                            Space.EMPTY,
+                            Collections.singletonList(JRightPadded.build((Statement) emptyBlock)),
+                            Markers.EMPTY);
+                    caseType = J.Case.Type.Rule;
+                }
+                else {
+                    paddedLabel = JRightPadded.build((J) defaultLabel);
+
+                    final J.Break breakStmt = new J.Break(
+                            Tree.randomId(),
+                            Space.build(NEW_LINE, Collections.emptyList()),
+                            Markers.EMPTY,
+                            null);
+                    statements = JContainer.build(
+                            Space.EMPTY,
+                            Collections.singletonList(JRightPadded.build((Statement) breakStmt)),
+                            Markers.EMPTY);
+                    caseType = J.Case.Type.Statement;
+                }
+
+                cases.add(new J.Case(
+                        Tree.randomId(),
+                        Space.build(NEW_LINE, Collections.emptyList()),
+                        Markers.EMPTY,
+                        caseType,
+                        JContainer.build(Space.EMPTY, Collections.singletonList(paddedLabel),
+                                Markers.EMPTY),
+                        statements,
+                        null,
+                        null));
+
+                visited = visited.withCases(casesBlock.withStatements(cases));
+                visited = autoFormat(visited, executionContext);
+
+                final Markers newMarkers = visited.getMarkers()
+                        .removeByType(CheckstyleViolationMarker.class);
+                visited = Optional.of(newMarkers).map(visited::withMarkers).get();
+            }
+
+            return visited;
+        }
+
+        private boolean isViolationMarked(J.Switch switchStmt) {
+            return switchStmt.getMarkers().findAll(CheckstyleViolationMarker.class).stream()
+                    .anyMatch(marker -> marker.isFor(CheckFullName.MISSING_SWITCH_DEFAULT));
+        }
+
+        private List<Statement> tryAddBreakToLastCase(List<Statement> cases) {
+            List<Statement> result = cases;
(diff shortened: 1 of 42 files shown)


MonuChaudhary14:
I am still working on this recipe

There are two mutation, which are left to be removed

MonuChaudhary14:
@rdiachenko 

Please review this PR

src/test/resources/org/checkstyle/autofix/recipe/missingswitchdefault/arrowstyle/OutputArrowStyle.java
```diff
@@ -0,0 +1,18 @@
...
+  </module>
+</module>
+*/
+package org.checkstyle.autofix.recipe.missingswitchdefault.arrowstyle;
+
+public class OutputArrowStyle {
+    void foo(int x) {
+        switch (x) {
+            case 1 -> System.out.println(1);
+            default->{
```
rdiachenko: missed spaces: `default->{` -> `default -> {`, please fix

MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,203 @@
...
+    private static final String NEW_LINE = "\n";
+    private static final String SPACE_STRING = " ";
+
+    private static final Set<Class<?>> TERMINAL_STATEMENTS =
+            new HashSet<>(Arrays.asList(
+                    J.Break.class, J.Return.class, J.Throw.class, J.Continue.class
+            ));
+
+    public MissingSwitchDefault() {
+    }
```
rdiachenko, line 54: this is redundant

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,203 @@
...
+
+        @Override
+        public J.Switch visitSwitch(J.Switch switchStmt, ExecutionContext executionContext) {
+            J.Switch visited = super.visitSwitch(switchStmt, executionContext);
+
+            if (isViolationMarked(visited)) {
+                final J.Block casesBlock = visited.getCases();
+                final List<Statement> cases = new ArrayList<>(casesBlock.getStatements());
+
+                final boolean isArrowStyle = isArrowStyleSwitch(cases);
```
rdiachenko: it is rather 'lambda' style than 'arrow'

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,203 @@
...
+
+                if (!isArrowStyle) {
+                    maybeAddBreakToLastCase(cases);
```
rdiachenko: `maybe` -> `try` - common naming in such cases

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,203 @@
...
+                            Markers.EMPTY,
+                            null);
+                    statements = JContainer.build(
+                            Space.EMPTY,
+                            Collections.singletonList(JRightPadded.build((Statement) breakStmt)),
+                            Markers.EMPTY);
+                }
+
+                final List<JRightPadded<J>> defaultLabelList;
+                if (isArrowStyle) {
```
rdiachenko: the logic is hard to follow and maintain, you have 3 conditions involving `isArrowStyle`, all spread around. please refactor this method and simplify

consider moving independent logical parts into their own methods

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,203 @@
...
+                if (needsBreak) {
+                    final J.Break prevBreak = new J.Break(
+                            Tree.randomId(),
+                            Space.build(NEW_LINE, Collections.emptyList()),
+                            Markers.EMPTY,
+                            null);
+                    lastCaseStatements.add(prevBreak);
+                    final J.Case updatedLastCase =
+                            lastCase.withStatements(lastCaseStatements);
+                    cases.set(cases.size() - 1, updatedLastCase);
```
rdiachenko: let's not mutate the method parameter, it is error prone on the caller side, since it needs to know that the provided list was mutated, which is a source of many bugs. instead, build a resulted list and return it if needed

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: now, we have to add them because of new checkstyle rules in 13.9.0

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: please check if this what you wanted

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: check this also

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,232 @@
...
+                final boolean isLambdaStyle = isLambdaStyleSwitch(cases);
+
+                if (!isLambdaStyle) {
+                    cases = tryAddBreakToLastCase(cases);
+                }
+
+                cases.add(createDefaultCase(isLambdaStyle));
+
+                visited = visited.withCases(casesBlock.withStatements(cases));
+                visited = maybeAutoFormat(switchStmt, visited, executionContext);
```
rdiachenko: `maybeAutoFormat` -> `tryAutoFormat`

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,232 @@
...
+                final J.Block casesBlock = visited.getCases();
+                List<Statement> cases = new ArrayList<>(casesBlock.getStatements());
+
+                final boolean isLambdaStyle = isLambdaStyleSwitch(cases);
+
+                if (!isLambdaStyle) {
+                    cases = tryAddBreakToLastCase(cases);
+                }
+
+                cases.add(createDefaultCase(isLambdaStyle));
```
rdiachenko: let's simplify, you don't need a separate `createDefaultCase` method for this. refactor to:
```
if (isLambdaStyleSwitch(cases)) {
 ...
 cases.add(lambda style)
} else {
 ...
  cases.add(statement style)
}
```

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,232 @@
...
+
+        private J.Case createLambdaStyleDefaultCase() {
+            final J.Identifier defaultLabel = new J.Identifier(
+                    Tree.randomId(),
+                    Space.EMPTY,
+                    Markers.EMPTY,
+                    null,
+                    DEFAULT_STRING,
+                    null,
+                    null);
```
rdiachenko: could you recheck if there is another constuctor that helps to avoiding passing `null` values. also I'd like to understand why we pass nulls if this is the only way to construct the object.

same question for below construction of `J.Case(...)`

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,232 @@
...
+
+            return new J.Case(
+                    Tree.randomId(),
+                    Space.build(NEW_LINE, Collections.emptyList()),
+                    Markers.EMPTY,
+                    J.Case.Type.Statement,
+                    JContainer.build(Space.EMPTY, defaultLabelList, Markers.EMPTY),
+                    statements,
+                    null,
+                    null);
```
rdiachenko: most of it is a duplicate of `createLambdaStyleDefaultCase()`. please think how to restructer code to reuse most common part

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,232 @@
...
+
+        private boolean isViolationMarked(J.Switch switchStmt) {
+            return switchStmt.getMarkers().findAll(CheckstyleViolationMarker.class).stream()
+                    .anyMatch(marker -> marker.isFor(CheckFullName.MISSING_SWITCH_DEFAULT));
+        }
+
+        private List<Statement> tryAddBreakToLastCase(List<Statement> cases) {
+            List<Statement> result = cases;
+            if (!cases.isEmpty()) {
+                final J.Case lastCase = (J.Case) cases.get(cases.size() - 1);
```
rdiachenko: i think you can use `getLast()` here, same for similar cases below

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,232 @@
...
+                            lastCase.withStatements(lastCaseStatements);
+                    updatedCases.set(updatedCases.size() - 1, updatedLastCase);
+                    result = updatedCases;
+                }
+            }
+            return result;
+        }
+
+        private boolean isLambdaStyleSwitch(List<Statement> cases) {
+            boolean lambdaStyle = false;
```
rdiachenko: rename to `result`

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
```diff
@@ -0,0 +1,232 @@
...
+            if (!cases.isEmpty()) {
+                lambdaStyle = ((J.Case) cases.get(0)).getType() == J.Case.Type.Rule;
```
rdiachenko: same api as mentioned above: `getLast() / getFirst()`, please try

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: No, there are no other constructor for both `J.Identifier` and `J.Case`
```
final J.Identifier defaultLabel = new J.Identifier(
    Tree.randomId(),
    Space.EMPTY,
    Markers.EMPTY,
    null,           // 1. List<J.Annotation> annotations
    DEFAULT_STRING,
    null,           // 2. JavaType type
    null            // 3. JavaType.Variable fieldType
);
```
annotations (1st null): Represents any annotations attached to the identifier (like @Deprecated). Since the default keyword doesn't have annotations, there is a null

type (2nd null): Represents the resolved semantic type of the identifier ( java.lang.String). Because default is a keyword and doesn't hold a semantic object type, there is a null

fieldType (3rd null): Represents metadata if the identifier refers to a specific class field. The default keyword is not a field, so this is null.

For `J.case`
```
cases.add(new J.Case(
    Tree.randomId(),
    Space.build(NEW_LINE, Collections.emptyList()),
    Markers.EMPTY,
    caseType,
    JContainer.build(Space.EMPTY, Collections.singletonList(paddedLabel), Markers.EMPTY),
    statements,
    null,           // 1. JContainer<Expression> body
    null            // 2. J.Case.Guard guard
));
```
body (1st null): In newer Java versions (Java 14+), we can have expressions directly yielded by the switch case (case A -> "Result";). Because we are assigning our logic to the statements parameter instead of a single expression body, this is null.

guard (2nd null): In Java 21+, we can have pattern matching guards (case Integer i when i > 0:). Because our default case doesn't have a when clause, the guard is null

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: please check the new code

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: done

src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java
MonuChaudhary14: done

rdiachenko (approved):
lgtm

PR #358, opened 2026-07-31
Issue #237: Removed the last unit test from ViolationMarkerRecipeTest

MonuChaudhary14:
Issue #237


src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -308,7 +308,8 @@ public J.CompilationUnit visitCompilationUnit(
                         result = result.withMarkers(result.getMarkers().add(marker));
                     }
                 }
-                result = result.withMarkers(result.getMarkers().add(APPLIED_MARKER));
+                final MarkersApplied markerToAdd = APPLIED_MARKER.withId(APPLIED_MARKER.getId());
```
sabaka: @MonuChaudhary14 could you explain what's going on here? Not sure I understand it.

MonuChaudhary14: When i removed the unit test, there are two mutations which are surviving

```
  <mutation unstable="false">
    <sourceFile>ViolationMarkerRecipe.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.marker.ViolationMarkerRecipe$MarkersApplied</mutatedClass>
    <mutatedMethod>getId</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.NullReturnValsMutator</mutator>
    <description>replaced return value with null for org/checkstyle/autofix/marker/ViolationMarkerRecipe$MarkersApplied::getId</description>
    <lineContent>return id;</lineContent>
  </mutation>

  <mutation unstable="false">
    <sourceFile>ViolationMarkerRecipe.java</sourceFile>
    <mutatedClass>org.checkstyle.autofix.marker.ViolationMarkerRecipe$MarkersApplied</mutatedClass>
    <mutatedMethod>withId</mutatedMethod>
    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.NullReturnValsMutator</mutator>
    <description>replaced return value with null for org/checkstyle/autofix/marker/ViolationMarkerRecipe$MarkersApplied::withId</description>
    <lineContent>return (M) new MarkersApplied(uuid);</lineContent>
  </mutation>
  ```
  
To kill these, i forced the recipe to use those methods
  
1. Killing the getId() Null Return Mutation:
Pitest modifies getId() so it always returns null

I changed the code to read the ID before adding the marker: 
`final MarkersApplied markerToAdd = APPLIED_MARKER.withId(APPLIED_MARKER.getId());`

If PITest mutates getId() to return null, the code effectively does APPLIED_MARKER.withId(null).
A marker with a null ID is attached to the tree
In AbstractRecipeTestSupport, the new validation runs:
`Assertions.assertNotNull(marker.getId(), "MarkersApplied ID must not be null");`
The assertion sees the null ID and fails the test and mutation Killed

2. Killing the withId() Null Return Mutation:
Pitest modifies withId() so it always returns null

If Pitest mutates withId() to return null, then the new code 
`APPLIED_MARKER.withId() resolves to null`
The recipe attempts to add null to the AST's markers list
In AbstractRecipeTestSupport, the new validation runs: 
`Assertions.assertTrue(hasMarkersApplied, "MarkersApplied must be present");`
Because null was added instead of the actual marker, the marker is missing from the tree
The assertion fails and mutation Killed

Comment (sabaka) [reply src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java]: Do you mind getting me a link to pit report before this changes please?

Comment (MonuChaudhary14) [reply src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java]: Should i open a dummy PR to show the pit report, after removing the unit test

Comment (MonuChaudhary14) [reply src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java]: Do you want this :
https://github.com/checkstyle/checkstyle-openrewrite-recipes/actions/runs/30440113562/job/90537139641

This is the pit report link for the previous commit: https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/351

**Comment by MonuChaudhary14** [comment]
@sabaka 

Ping

File: src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java (hunk 1 above)
Comment (rdiachenko) [reply src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java]: @MonuChaudhary14 the changes in this PR look like a hack. We need a proper fix. Please think whether we actually need `MarkersApplied`, and what it takes to simplify the logic by removing it?

Comment (MonuChaudhary14) [reply src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java]: @rdiachenko 

I have two ways, 
First one:
I tried to make MarkerVisitor inherently idempotent by checking individual nodes. Before adding a CheckstyleViolationMarker to an AST node (like J.Switch or J.Import), the recipe would check if that specific node already had a violation marker.

But it failed, because OpenRewrite runs multiple recipes in a cycle. ViolationMarkerRecipe runs first, adding markers to nodes. Then, other downstream recipes run. Recipes like UseEnhancedSwitch often replace entire AST nodes with new ones (e.g., replacing J.Switch with J.SwitchExpression). When OpenRewrite starts Cycle 2, our recipe runs again. It inspects the new J.SwitchExpression node, sees that it lacks a CheckstyleViolationMarker (because it's a new node created by another recipe), and re-adds the marker. This causes a change in Cycle 2, failing the test.

Second one:
Since ViolationMarkerRecipe extends ScanningRecipe<Accumulator>, I tried to store the state in the accumulator itself. I added a Set<Path> processedFiles to the Accumulator. In Cycle 1, I add the file to the set. In Cycle 2, I check the set and skip processing

It also failed because:
Between cycles, OpenRewrite isolates state by calling clone() on the ScanningRecipe
The clone() implementation generates a brand new UUID for the recipe's accumulator message key. When Cycle 2 starts, getAccumulator() is called. Because the UUID changed, it creates a brand new, empty Accumulator. The state from Cycle 1 is lost, and the recipe re-adds all markers, causing a 2-cycle failure

Comment (sabaka) [reply src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java]: @MonuChaudhary14 @rdiachenko well, one of the obvious options would be to make the record package private and test independently the same way as `CheckstyleViolationMarker` is tested just to get unblocked. Also isn't great, but doesn't pollute the logic with weird id swaps at least.
Alternatively, can we get rid of the marker and store something in the execution context? Something like this:
```java
         private static final String MARKED_FILES =
            "org.checkstyle.autofix.marker.ViolationMarkerRecipe.markedFiles";
        @Override
        public J.CompilationUnit visitCompilationUnit(
                J.CompilationUnit compUnit, ExecutionContext executionContext) {
            final Set<Path> markedFiles = executionContext.computeMessageIfAbsent(
                    MARKED_FILES, key -> ConcurrentHashMap.newKeySet());

            J.CompilationUnit result = compUnit;
            if (markedFiles.add(compUnit.getSourcePath())) {
                fileMarkers = acc.getByFile(compUnit.getSourcePath());
                result = super.visitCompilationUnit(compUnit, executionContext);

                final List<CheckstyleViolationMarker> markers = fileMarkers.get(result.getId());
                if (markers != null) {
                    for (CheckstyleViolationMarker marker : markers) {
                        result = result.withMarkers(result.getMarkers().add(marker));
                    }
                }
            }
            return result;
        }
```
Or this won't work?

MonuChaudhary14: @rdiachenko @sabaka 

I found a solution. `Please let me know whether this approach will work for you`

Currently, the code is modifying the LST by attaching the invisible `MarkersApplied` object to files, just to remember that ViolationMarkerRecipe had already processed them.

New Approach:
We can have ViolationMarkerRecipe remember which files it has seen in memory, without touching the LST

```
private final Set<Path> processedFiles = ConcurrentHashMap.newKeySet();
```

When the recipe processes a file in Cycle 1, it adds the file's path to this Set. If OpenRewrite triggers Cycle 2 and asks the recipe to process the same file again, the recipe checks the Set, sees the file is already there, and immediately skips it

While tesing, i found that , OpenRewrite's RecipeScheduler was calling CheckstyleAutoFix.getRecipeList() at the beginning of every single cycle

Because it returned a brand new instance of ViolationMarkerRecipe on Cycle 2, that new instance will have a completely empty processedFiles Set.It had forgotten all the files it processed in Cycle 1, so it processed them all over again, modifying the LST and causing an infinite loop

To solve this, we can introduce, cachedRecipeList to CheckstyleAutoFix. We can change the method so it only creates the recipes once

In Cycle 2, OpenRewrite gets the exact same ViolationMarkerRecipe instance that ran in Cycle 1
Because it is the same instance, its processedFiles Set is still intact

It successfully remembers the files it already processed, skips them, and the cycle safely stabilizes

As a massive bonus, we completely eliminated the performance penalty of repeatedly re-reading and re-parsing the Checkstyle XML report from the hard drive on every single cycle

MonuChaudhary14: @sabaka 

I think mine is similar to yours, but you have made it much cleaner by adding the set in ExecutionContext

i will soon provide the changes

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -135,14 +145,16 @@ private void processFileViolations(J.CompilationUnit compilationUnit, Path sourc
             printer.visit(compilationUnit, capture.append(""), rootCursor);
 
             final Map<UUID, List<CheckstyleViolationMarker>> markersToAdd = new HashMap<>();
-            for (CheckstyleViolation violation : fileViolations) {
+            final Iterator<CheckstyleViolation> iterator = fileViolations.iterator();
+            do {
+                final CheckstyleViolation violation = iterator.next();
```
illialean, line 150: Why expicit iterator over foreach?

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -112,13 +114,21 @@ private final class ScannerVisitor extends JavaIsoVisitor<ExecutionContext> {
         @Override
         public J.CompilationUnit visitCompilationUnit(J.CompilationUnit compUnit,
                                                       ExecutionContext executionContext) {
+            final Set<Path> processedFiles = executionContext.computeMessageIfAbsent(
+                    MARKED_FILES, key -> ConcurrentHashMap.newKeySet());
             final Path sourcePath = compUnit.getSourcePath();
-            final List<CheckstyleViolation> fileViolations = violations.stream()
-                    .filter(violation -> violation.getFilePath().endsWith(sourcePath))
-                    .toList();
+            final J.CompilationUnit result = compUnit;
```
illialean: you don't really need this variable, it's the same as comUnit without any modifications

illialean (commented):
Other than iterator and unnecessary variable looks good

sabaka:
Sorry, sent comments from the wrong account 🤦

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
MonuChaudhary14: This was done to avoid the mutation
In this we are not checking `isEmpty()` and calls `iterator.next()` unconditionally at least once any mutation that removes if (!fileViolations.isEmpty()) check will cause to throw a NoSuchElementException

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
MonuChaudhary14: thanks for pointing out, i have removedn that

Issue #365, opened 2026-08-04
Implement recipe for AvoidStaticImport


MonuChaudhary14:
# Description
Check: hhttps://checkstyle.sourceforge.io/checks/imports/avoidstaticimport.html
Check core OpenRewrite repository and reuse existing recipes if present.
Update README with the recipe link and notes about coverage if needed.

PR #366, opened 2026-08-04
Issue #365: Implemented the recipe for AvoidStaticImport

MonuChaudhary14:
Issue #365


MonuChaudhary14:
There is still a changes needed, there are a lot of bugs, which are causing the kafka report to fail

PR #374, opened 2026-08-10
Issue #376: Refractored CheckstyleViolationMarker to use specific typed markers

MonuChaudhary14:
Issue #376


MonuChaudhary14:
@rdiachenko 
Please check this

rdiachenko (commented):
1. Let's convert `CheckstyleViolationMarker` to interface, something like this:
```java
  public interface CheckstyleViolationMarker extends Marker {

      UUID id();

      CheckstyleViolation violation();

      @Override
      default UUID getId() {
          return id();
      }

      @Override
      @SuppressWarnings("unchecked")
      default <M extends Marker> M withId(UUID newId) {
          return (M) MarkerRegistry.create(newId, violation());
      }
  }
```

2. Every check marker becomes:
```java
  public record FinalLocalVariableMarker(UUID id, CheckstyleViolation violation)
          implements CheckstyleViolationMarker {
  }
```

3. Let's move `MARKER_FACTORIES` related logic into `MarkerRegistry` as registry and reuse it in both `ViolationMarkerRecipe` and the default `withId` above:
```java
  private static final Map<CheckFullName, BiFunction<UUID, CheckstyleViolation,
          CheckstyleViolationMarker>> FACTORIES = new EnumMap<>(CheckFullName.class);
```

4. Yes, recipes become one liners:
```java
return variable.getMarkers().findFirst(FinalLocalVariableMarker.class).isPresent();
```

**Note:** if it's easier, let's apply this for all recipes as a single atomic change, if the migration and split onto multiple PRs creates more work.

MonuChaudhary14:
@rdiachenko 

Please check these changes
i am also trying to kill the marker mutations, will update you soon

README.md
```diff
@@ -147,7 +147,7 @@ This table tracks the auto-fix support status of OpenRewrite recipes for each Ch
 | 🔴     | [`MatchXpath`](https://checkstyle.sourceforge.io/checks/coding/matchxpath.html#MatchXpath)                               |           | Context-dependent XPath violations |
 | 🔴     | [`MissingCtor`](https://checkstyle.sourceforge.io/checks/coding/missingctor.html#MissingCtor)                             |           | Add explicit constructor |
 | 🟢     | [`MissingNullCaseInSwitch`](https://checkstyle.sourceforge.io/checks/coding/missingnullcaseinswitch.html#MissingNullCaseInSwitch) | `TBD`           |                |
-| 🟢     | [`MissingSwitchDefault`](https://checkstyle.sourceforge.io/checks/coding/missingswitchdefault.html#MissingSwitchDefault)   | `org.checkstyle.autofix.recipe.MissingSwitchDefault` |                |
+| 🟢     | [`MissingSwitchDefault`](https://checkstyle.sourceforge.io/checks/coding/missingswitchdefault.html#MissingSwitchDefault)   | [`MissingSwitchDefault`](https://github.com/checkstyle/checkstyle-openrewrite-recipes/blob/main/src/main/java/org/checkstyle/autofix/recipe/MissingSwitchDefault.java) |                |
```
MonuChaudhary14, line 150: In the previous recipe implementation (MissingSwitchDefault), i by mistake added the wrong test in url
This is the reason for this change

MonuChaudhary14:
@rdiachenko 

This PR is done, you can check it now

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
```diff
@@ -153,14 +154,20 @@ private Map<UUID, List<CheckstyleViolationMarker>> findSmallestNode(
...
+                                new ArrayList<>(List.of(marker));
                         return Map.of(targetId, markerList);
                     })
                     .orElseGet(Collections::emptyMap);
         }
 
+        private CheckstyleViolationMarker createMarker(CheckstyleViolation violation) {
+            final CheckstyleViolationMarker marker =
+                    MarkerRegistry.create(Tree.randomId(), violation);
+            return marker.withId(marker.getId());
```
rdiachenko: why do we need to call `withId` and can't return just `marker`?

config/pitest-recipe-2-suppressions.xml
```diff
@@ -1,12 +1,3 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <suppressedMutations>
-  <!-- Cannot be killed by integration tests until a second recipe lands a marker on this node type. See issue #236 -->
-  <mutation unstable="false">
-    <sourceFile>AnnotationOnSameLine.java</sourceFile>
-    <mutatedClass>org.checkstyle.autofix.recipe.AnnotationOnSameLine$AnnotationModifierVisitor</mutatedClass>
-    <mutatedMethod>lambda$hasMarker$0</mutatedMethod>
-    <mutator>org.pitest.mutationtest.engine.gregor.mutators.returns.BooleanTrueReturnValsMutator</mutator>
-    <description>replaced boolean return with true for org/checkstyle/autofix/recipe/AnnotationOnSameLine$AnnotationModifierVisitor::lambda$hasMarker$0</description>
-    <lineContent>.anyMatch(marker -&gt; marker.isFor(CheckFullName.ANNOTATION_ON_SAME_LINE));</lineContent>
```
rdiachenko, line 10: could you confirm we fixed all such mutations by this approach?

rdiachenko (commented):
nice! we still have duplication, since each new recipe now requires a corresponding marker type, but it a very good step forward to kill mutations and not depend on multiple recipes covering same node to kill it.

few notes below

src/main/java/org/checkstyle/autofix/marker/ViolationMarkerRecipe.java
MonuChaudhary14: thank you for pointing out, i updated it with 

```
        private CheckstyleViolationMarker createMarker(CheckstyleViolation violation) {
            return MarkerRegistry.create(Tree.randomId(), violation);
        }
```

config/pitest-recipe-2-suppressions.xml
MonuChaudhary14: yes, all the mutations are killed now, with these

for this you can check 

[src/test/resources/org/checkstyle/autofix/recipe/emptystatement/illegaltoken/DiffIllegalToken.diff](https://github.com/checkstyle/checkstyle-openrewrite-recipes/pull/374/changes#diff-f42dd515649aa61e5ff2c5679baed86d9dbc4c5022c5ff58c6c0ca981475d3faL1)

i have deleted this file, because it uses illegalToken recipe, used to kill. the mutation of EmptyStatement

rdiachenko (approved):
lgtm

